PR #12, opened 2015-01-10
http bits of the monster that was http2.0 implementation

bryce-anderson:
This implements the HTTP/2.0 protocol in the form of codecs and a server pipeline stage.

Any suggestions anybody has would be nice.


examples/src/main/scala/org/http4s/blaze/examples/NIO2HttpServer.scala
```diff
@@ -12,10 +12,10 @@ import scala.concurrent.duration._
 
 class NIO2HttpServer(port: Int) {
 
-  val group = AsynchronousChannelGroup.withFixedThreadPool(10, java.util.concurrent.Executors.defaultThreadFactory())
+  val group = AsynchronousChannelGroup.withFixedThreadPool(6, java.util.concurrent.Executors.defaultThreadFactory())
```
rossabaker: Would using some multiple of the available processors from `Runtime` be appropriate here?

examples/src/main/scala/org/http4s/blaze/examples/SSLHttpServer.scala
```diff
@@ -16,13 +17,10 @@ class SSLHttpServer(port: Int) {
...
     eng.setUseClientMode(false)
-
-    TrunkBuilder(new SSLStage(eng, 100*1024)).cap(new ExampleHttpServerStage(None, 10*1024))
+    TrunkBuilder(new SSLStage(eng, 100*1024)).cap(ExampleService.http1Stage(None, 10*1024))
   }
 
-  val group = AsynchronousChannelGroup.withFixedThreadPool(10, java.util.concurrent.Executors.defaultThreadFactory())
-
-  private val factory = new NIO2SocketServerChannelFactory(f)
+  private val factory = new NIO1SocketServerChannelFactory(f, workerThreads = 6)
```
rossabaker: Here's that magic number 6 again.

examples/src/main/scala/org/http4s/blaze/examples/WebSocketServer.scala
```diff
@@ -62,4 +54,18 @@ class SocketStage extends WSStage {
...
+  private val logger = getLogger
+
+  def handleRequest(method: Method, uri: Uri, headers: Seq[(String, String)], body: ByteBuffer): Future[Response] = {
+    if (WebsocketHandshake.isWebSocketRequest(headers)) {
+      logger.info(s"Received a websocket request at $uri")
+
+      // Note the use of WSStage.segment. This makes a pipeline segment that includes a serializer so we
+      // can safely write as many messages as we want without worrying about clashing with pending writes
+      Future.successful(WSResponse(WSStage.bufferingSegment(new SocketStage)))
+    } else Future.successful(SimpleHttpResponse.Ok("Use a websocket!\n" + uri))
```
rossabaker: Ok?

examples/src/main/scala/org/http4s/blaze/examples/NIO2HttpServer.scala
rossabaker: Oh, as I started to complain about these sixes elsewhere, I see it's examples.  But, still.

http/src/main/scala/org/http4s/blaze/http/http20/BasicHttpStage.scala
```diff
@@ -0,0 +1,144 @@
...
+      logger.debug("EOF while accumulating body")
+      shutdownWithCommand(Cmd.Disconnect)
+
+    case Failure(t) =>
+      logger.error(t)("Error in getBody(). Headers: " + hs)
+      val e = INTERNAL_ERROR()
+      shutdownWithCommand(Cmd.Error(e))
+  }
+
+  private def checkAndRunReqeust(hs: Headers, body: ByteBuffer): Unit = {
```
rossabaker: It's private, but spelling: request

http/src/main/scala/org/http4s/blaze/http/http20/ByteBufferInputStream.scala
```diff
@@ -0,0 +1,42 @@
...
+import java.io.{IOException, InputStream}
+import java.nio.ByteBuffer
+
+final class ByteBufferInputStream(buffer: ByteBuffer) extends InputStream {
+
+  private var markSize = 0
+
+  override def read(): Int = {
+    markSize -= 1
+    buffer.get()
```
rossabaker: Don't you need to do a remaining check?

Also, `buffer.get` returns a signed `byte`, if I remember correctly.  You need to do some bit math to get this into the [0, 256) range.

http/src/main/scala/org/http4s/blaze/http/http20/Http20FrameDecoder.scala
```diff
@@ -0,0 +1,280 @@
...
+        
+        // this concludes the types established by HTTP/2.0, but it could be an extension
+      case code                     => onExtensionFrame(code, streamId, flags, buffer.slice())
+    }
+
+    // reset buffer limits
+    buffer.limit(oldLimit)
+    buffer.position(endOfFrame)
+
+    return r
```
rossabaker: re-what now?

http/src/main/scala/org/http4s/blaze/http/http20/Http20FrameDecoder.scala
```diff
@@ -0,0 +1,280 @@
...
+        Error(PROTOCOL_ERROR(s"Padding, $padding, exceeds payload length: ${buffer.remaining}"))
+      }
+      else {
+        buffer.limit(buffer.limit() - padding)
+        Continue
+      }
+    }
+    else Continue
+  }
+}
```
rossabaker: // bookmark comment: @rossabaker made it to here before he ran out of time.

http/src/main/scala/org/http4s/blaze/http/http20/ByteBufferInputStream.scala
bryce-anderson: Yes, your right. Fortunately (maybe unfortunately) the twitter header decoder, which requires an ```InputStream```, always checks if data is available before reading, and converts the result back to a byte, so it avoids the [0,256) concern. Better to fix it in case they change their impl, and good catch.
This class will also be made private. There are numerous better solutions out there if someone needs one; I didn't want to pull in a dep for just this.

One such trick to get the unsigned range is ```buffer.get() & 0xff```

http/src/main/scala/org/http4s/blaze/http/http20/Http20FrameDecoder.scala
bryce-anderson: Is the concern that its a ```return``` statement at the end of the block instead of the expression form? I kept it return just to be consistent with the preceding code: this is really java written in scala. All the checks would have made this a rats nest of ```if/else``` statements. Perhaps it could be broken down into multiple functions, but I thought it nice to keep the code that mutates the buffer as localized as possible.

http/src/main/scala/org/http4s/blaze/http/http20/Http2ServerHubStage.scala
```diff
@@ -0,0 +1,517 @@
...
+  private implicit def ec = trampoline
+
+  type Http2Msg = NodeMsg.Http2Msg[HType]
+
+  override type Key = Int
+  override type Out = Http2Msg
+  override protected type Attachment = FlowControl.NodeState
+
+  // Using synchronization and this is the lock
+  private val lock = new AnyRef
```
rossabaker: What kind of contention are we expecting here?  Was there no atomic solution?

http/src/main/scala/org/http4s/blaze/http/http20/bits.scala
```diff
@@ -0,0 +1,52 @@
+package org.http4s.blaze.http.http20
+
+object bits {
+
+  def HeaderSize = 9
+  def clientTLSHandshakeString = "PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
```
rossabaker: Not going to make much difference, but these look like vals.

http/src/main/scala/org/http4s/blaze/http/http20/bits.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  object Masks {
+    val STREAMID = 0x7fffffff
+    val LENGTH   =   0xffffff
+    val int31    = 0x7fffffff
+    val exclsive = ~int31
```
rossabaker: typo?

http/src/main/scala/org/http4s/blaze/http/package.scala
```diff
@@ -27,7 +35,7 @@ package object http {
       SimpleHttpResponse("OK", 200, headers, ByteBuffer.wrap(body))
 
     def Ok(body: String, headers: Headers): SimpleHttpResponse =
-      Ok(body.getBytes(UTF_8), ("Content-Type", "text/plain; charset=UTF-8")+:headers)
+      Ok(body.getBytes(UTF_8), ("content-type", "text/plain; charset=utf-8")+:headers)
```
rossabaker: Why is this lower case?  It's insensitive, but this seems ... non-canonical.  I only ask because I wonder if I missed something important.

http/src/main/scala/org/http4s/blaze/http/websocket/WSFrameAggregator.scala
```diff
@@ -15,7 +15,7 @@ class WSFrameAggregator extends MidStage[WebSocketFrame, WebSocketFrame] {
 
   def name: String = "WebSocket Frame Aggregator"
 
-  private val queue = new ListBuffer[WebSocketFrame]
+  private var queue = new ArrayBuffer[WebSocketFrame]
```
rossabaker: When you need a mutable reference to a mutable value, you know you're in the shit.

http/src/main/scala/org/http4s/blaze/http/http20/Http20FrameDecoder.scala
rossabaker: More of a kneejerk reaction, and a deep production scar from how `return` interacts with local exception handling.  (I try to minimize both ingredients these days.)  Inner methods tend to do a good job of containing scope without relying on `return`, but mileage varies on whether people find that more or less readable.

http/src/main/scala/org/http4s/blaze/http/package.scala
bryce-anderson: http2 requires all header names to be lower case. Its [malformed otherwise](http://http2.github.io/http2-spec/index.html#HttpHeaders). I believe I make sure they are lowercase elsewhere, but why make a new string.

http/src/main/scala/org/http4s/blaze/http/http20/Http2ServerHubStage.scala
bryce-anderson: Events from streams modify things like flow control windows, the header compressor state, making and closing streams etc. Some events depend on many of the different parts of state. The only way to maintain all of it atomically would be to have a super state type which would be swapped out atomically, or technically an actor is based on atomic operations.

http/src/main/scala/org/http4s/blaze/http/package.scala
rossabaker: Heh. That indicates the quality of review you received. :-)

PR #16, opened 2015-03-25
Accept a bare LF in the status line and headers

bryce-anderson:
Associated with https://github.com/http4s/http4s/issues/236

This should be a no brainer: the old spec (http://tools.ietf.org/html/rfc2616#section-19.3) recommends allowing LF without the preceeding CR. Note I didn't see this in rfc 7230 which replaces 2616, but I believe it should still apply.

I'll merge this shortly unless there is objection.


rossabaker:
:+1:

PR #19, opened 2015-04-11
Topic/concrete hs

bryce-anderson:
I think this is good. The only cost is the translation of ````Seq[(String,String)]```` to whatever type of collection the DSL wants. The concrete type of the collection used by default in this package is an ```ArrayBuffer[(String,String)]``` so it should be reasonably GC friendly.

If nobody has any comments in a few days, I'll just merge this.


http/src/main/scala/org/http4s/blaze/http/http20/Http2StageTools.scala
```diff
@@ -0,0 +1,31 @@
...
+  val TE = "te"
+  val ContentLength = "content-length"
+
+  def isLowerCase(str: String): Boolean = {
+    @tailrec
+    def go(i: Int): Boolean = {
+      if (i == str.length) true
+      else {
+        val ch = str.charAt(i)
+        if ('A' <= ch && ch <= 'Z') false
```
rossabaker: What about diacritics, or are those not valid here?

rossabaker:
I'm not deeply up to speed on the architecture here, but it seems reasonable.

http/src/main/scala/org/http4s/blaze/http/http20/Http2StageTools.scala
bryce-anderson: to be very strict I should check to make sure its lower case ascii. This was done just so I could match against the special headers as Strings.

bryce-anderson:
Thanks for the comments.

PR #20, opened 2015-05-08
Handle the error associated with opening a connection

bryce-anderson:
Nonfatal exceptions are caught and used to complete the Future with a
failure.


rossabaker:
:+1:

PR #21, opened 2015-06-03
Big changes to connection initialization and lifetime management.

bryce-anderson:
__NIO1 and NIO2 now both understand a thread pool__

The resource pool, or NIOXSocketServerGroup, becomes a shared resource. With if you can `.bind(address,service)` and get a ServerChannel. The Server channel can be canceled, but the NIOXSocketServerGroup remains intact.

This makes it simple to easily spin up and down new server services which share a group of threads.

__Change shutdown hook registration behavior__

`addShutdownHook` now returns a Boolean where true signifies successful hook registration.

__Also consolidated examples__

__Future work__
In the future, NIO1 groups may eliminate their connection accepter thread and instead use the SelectorPool threads to accept connections.


bryce-anderson:
Any review on this would be greatly appreciated. Its a lot of changes, and I know I'm by far the most familiar with this code base, but a second set of eyes is very useful.

bryce-anderson:
Also, for motivation see https://github.com/http4s/http4s/issues/197.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
```diff
@@ -0,0 +1,238 @@
...
+              // check to see if we want to keep this connection
+              if (acceptConnection(address)) {
+                clientChannel.setOption(java.net.StandardSocketOptions.TCP_NODELAY, java.lang.Boolean.FALSE)
+                loop.initChannel(service, clientChannel, key => new SocketChannelHead(clientChannel, loop, key))
+              }
+              else clientChannel.close()
+            }
+          }
+          catch {
+            case s: SecurityException => logger.info(s)("Connection rejected by SecurityManager.")
```
rossabaker: Just info?

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
```diff
@@ -0,0 +1,238 @@
...
+
+      }
+
+      // Close down the selector loops
+      pool.shutdown()
+
+      // clear out the queue
+      queue.getAndSet(null).foreach { ch =>
+        try ch.close()
+        catch { case NonFatal(_) => /* NOOP */ }
```
rossabaker: I might log.debug these as a compromise between hiding harmless exceptions and being wrong about what's harmless.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
bryce-anderson: That seemed appropriate at the time. Does this warrant a ```warn```?

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
bryce-anderson: :+1:

rossabaker:
https://twitter.com/iamdevloper/status/397664295875805184

I would bet we both missed things, but, looks like progress to me.  :+1:

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
rossabaker: Did it do what was asked of it?  If not, I consider that an error.

bryce-anderson:
Hehe, that tweet made the rounds through my feed a couple weeks ago and I thought about it before asking for this code review. Thanks for taking a look.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
bryce-anderson: This one is a bit of a fuzzy case.
```
SecurityException - If a security manager has been installed and it does not permit access to the remote endpoint of the new connection
```
This sounds like a situation where the endpoint is rejected based on being black-listed in some manner. To me that doesn't really seem like an error, but with the NIO1 api, there isn't a good way to express the rejection besides throwing the exception. 

I'll see about finding some prior art on this one.

bryce-anderson: Looking through source of Jetty, Netty, and Tomcat, I don't find any special treatment of this exception in the context of accepting connections. Therefor I believe you're right: this should be treated as an error.
Good catch.

PR #25, opened 2015-07-16
Fix QuietTimeoutStage so that it will cancel unused resets after its shutdown

bryce-anderson:
There was a condition where the tail performs a read/write request, and before it is honored, the pipeline gets shutdown (maybe by a disconnect). That disconnect shuts down the pipeline and cancels any pending timeouts. However, the pending IO request _should_ eventually resolve, and that will reset the stage again, scheduling another timeout.

This PR addresses this by flagging the atomic reference with a 'tag' which it checks when resetting the timeout and acts appropriately.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/pipeline/stages/TimeoutStageBase.scala
+++ b/core/src/main/scala/org/http4s/blaze/pipeline/stages/TimeoutStageBase.scala
@@ -1,15 +1,17 @@
 package org.http4s.blaze.pipeline.stages
 
+import scala.annotation.tailrec
 import scala.concurrent.Future
 import scala.concurrent.duration.Duration
 import org.http4s.blaze.pipeline.MidStage
 import org.http4s.blaze.pipeline.Command.{ Disconnect, Disconnected }
 import org.http4s.blaze.util.{ Cancellable, TickWheelExecutor }
 
-import org.http4s.blaze.util.Execution.scheduler
 import java.util.concurrent.atomic.AtomicReference
 
 abstract class TimeoutStageBase[T](timeout: Duration, exec: TickWheelExecutor) extends MidStage[T, T] { stage =>
+
+  import TimeoutStageBase.closedTag
 
   // Constructor
   require(timeout.isFinite() && timeout.toMillis != 0, s"Invalid timeout: $timeout")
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/http4s/blaze/pipeline/stages/TimeoutStageBase.scala
```diff
@@ -22,14 +24,21 @@ abstract class TimeoutStageBase[T](timeout: Duration, exec: TickWheelExecutor) e
...
     }
   }
 
-  protected def setAndCancel(next: Cancellable): Unit = {
+  @tailrec
+  final def setAndCancel(next: Cancellable): Unit = {
     val prev = lastTimeout.getAndSet(next)
-    if (prev != null) prev.cancel()
+    if (prev == closedTag && next != closedTag) {
+      // woops... we submitted a new cancellation when we were closed!
```
rossabaker: I don't have a whole lot to add, but every time I see a "woops" comment, I'm inclined to look for a log.warn so maybe the developer can see where they woopsed.

core/src/main/scala/org/http4s/blaze/pipeline/stages/TimeoutStageBase.scala
```diff
@@ -52,6 +61,11 @@ abstract class TimeoutStageBase[T](timeout: Duration, exec: TickWheelExecutor) e
   final protected def cancelTimeout(): Unit = setAndCancel(null)
 
   final protected def startTimeout(): Unit = resetTimeout()
-
-
 }
+
+private object TimeoutStageBase {
+  // Represents the situation where the pipeline has been closed.
+  val closedTag = new Cancellable {
```
rossabaker: Is this intentionally not using `Cancellable.noopCancel`?  Is it the reference equality check in `setAndCancel`?  That should work, but, ick.  How about a case object that's Cancellable?

core/src/main/scala/org/http4s/blaze/pipeline/stages/TimeoutStageBase.scala
bryce-anderson: This event is not always avoidable. For example (happens in http4s), the server pipeline attempts a read then the client disconnects. A `Disconnected` signal shuts down the pipeline including the timeout stage, but the pending read will also complete and the timeout stage will attempt to reset.

core/src/main/scala/org/http4s/blaze/pipeline/stages/TimeoutStageBase.scala
bryce-anderson: It is a reference equality which should be indistinguishable for a singleton defined by a val or a case object.

I didn't use `Cancellable.noopCancel` because the purpose is more refined: it serves as a tag, not just as a noop. Additionally, making it private in the case object keeps the meaning local so future contributors (and myself) won't need to think about what `noopCancel` does, who else uses it, could it leak into the method call by some other use, etc.  
Although now that you mention i I will use `noopCancel` to replace the `null` references.

PR #28, opened 2015-09-18
Add ability to splice stages before a Tail stage

bryce-anderson:
Added tests and fixed the broken spliceBefore on Head stage.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
+++ b/core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
@@ -43,7 +43,7 @@ sealed trait Stage {
     * before this method is called. It is not impossible for this method to be called multiple
     * times by misbehaving stages. It is therefore recommended that the method be idempotent.
     */
-  protected def stageStartup(): Unit = logger.debug(s"${getClass.getName} starting up at ${new Date}")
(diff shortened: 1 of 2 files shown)


bryce-anderson:
I need this for the http4s blaze client refactoring work: I want to be able to splice timeout stages into the pipeline on startup.

I plan to merge this rather fast so speak up if you have objections.

core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
```diff
@@ -314,17 +322,21 @@ trait MidStage[I, O] extends Tail[I] with Head[O] {
...
-    me._nextStage._prevStage = me._prevStage
-
-    _nextStage = null
-    _prevStage = null
-    this
+      _nextStage = null
+      _prevStage = null
+    }
+    else {
+      val e = new Exception("Cannot remove a disconnected stage")
```
rossabaker: I'm not sure this needs to be an exception: the end result is the stage isn't there.  I'd long a stern warning.

rossabaker:
:+1:.  I defer on the exception.

core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
bryce-anderson: Thats a good point. Its now a warning.

bryce-anderson:
Thanks for the review!

PR #32, opened 2015-12-09
Add a lenient parsing mode for headers

clayrat:
I've tried to keep the changes to the minimum. I use a new exception (subclassed from `BadRequest`) to signal a bad symbol, and a new state in `parseHeaders` to skip until the end of the current header entry. For this I had to split `next()` to do reading and validation separately.

I guess you might also want some tests to go with this, but I'm not sure where to put them.

### DIFF
--- a/http/src/main/java/org/http4s/blaze/http/http_parser/BaseExceptions.java
+++ b/http/src/main/java/org/http4s/blaze/http/http_parser/BaseExceptions.java
@@ -22,6 +22,12 @@ public class BaseExceptions {
         }
     }
 
+    public static class BadCharacter extends BadRequest {
+        public BadCharacter(String msg) {
+            super(msg);
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


bryce-anderson:
So as discussed in the http4s gitter channel, I think people like this feature but the preferred solution would be to emit the headers with bad chars as opposed to this PR's behavior of suppressing them. The downside is that it will take some more work to get that to happen because right now the illegal char `0x0` is used as a signalling value for the parser. We talked a bit about refactoring to use `0xFFFF` because that will never be found in a single byte from the input.

bryce-anderson:
@clayrat, are you interested in modifying this PR to the behaviour described above?

clayrat:
I can take a look, but probably not right away; maybe end of this week or later.

bryce-anderson:
Closed by #36. Thanks @clayrat!

PR #33, opened 2015-12-24
Code review changes, mostly just comments

hvesalai:
@bryce-anderson just say if you want me to rework the pull request before you merge it.


bryce-anderson:
Travis says there are problems getting the thread names in the asserts: I believe it should be `Thread.getName`, not `Thread.name`. I would prefer those be comments anyway and we make cached `_stopReadTask` and `_stopWriteTask`'s that all go through `pool.executeTask`. I think a method or two might lose utility at that point and can be culled.

Thanks again for the review, I'm glad there was nothing of note. :smile:

hvesalai:
@bryce-anderson commented out asserts, added @inline annotation to executeTask

PR #34, opened 2015-12-25
Content-Length and Transfer-Encoding handling

hvesalai:
Code-review changes with intent to make the code one step closer to what the specification mandates.


bryce-anderson:
See my comment for commit 3892cda for the details but in general, :+1:. 

By chance have you looked into what effect (if any) this will have on http4s proper? I doubt it will be significant but it might be worth double checking. I'll try to take a peak in the near future but probably not tonight.

hvesalai:
@bryce-anderson can you look at the test log. There are some exceptions in it although all tests should pass now. Are they expected?

hvesalai:
Ref to http4s proper. I think it is using the parser, so the content-length parsing changes should have an effect.

bryce-anderson:
@hvesalai, I believe those exceptions are expected. I suppose they could probably be replaced by a more consistent exception.

bryce-anderson:
@hvesalai, thanks for the PR!

PR #35, opened 2015-12-26
Topic/nio1 actorish

bryce-anderson:
This PR reformulates the `NIO1HeadStage` to run all mutation operations inside the `SelectorLoop` thread. I think it makes it a bit clearer what is happening and performance stays the same or perhaps better.

cc: @hvesalai


hvesalai:
LGTM

PR #36, opened 2015-12-26
Topic/lenient parsers

bryce-anderson:
This adds the ability to parse the http1 request prelude in a more 'lenient' fashion.

One notable difference here is that leniency is applied to all characters, not just those in header fields. I don't think this is a good thing. I'm also not exactly convinced that its a good idea at all to accept illegal characters so this one may not make the cut.

See also #32
cc @clayrat

### DIFF
--- a/http/src/main/java/org/http4s/blaze/http/http_parser/BaseExceptions.java
+++ b/http/src/main/java/org/http4s/blaze/http/http_parser/BaseExceptions.java
@@ -22,6 +22,12 @@ public class BaseExceptions {
         }
     }
 
+    public static class BadCharacter extends BadRequest {
+        public BadCharacter(String msg) {
+            super(msg);
+        }
+    }
+
     public static class BadResponse extends ParserException {
         public BadResponse(String msg) {
             super(msg);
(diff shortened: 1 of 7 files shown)


clayrat:
Yes, maybe it would make sense to limit this just to the headers, or introduce some level of "lenientness".

bryce-anderson:
@clayrat, one solution would be to bump the `isLenient` up a level and make it a parameter of the `next` function. All calls to next ouside of headers would simply pass false while in the headers blocks it could decide based on the parametrization of the parser. 

Before merging this (or related PR's) at all I think its a good idea to look into the security considerations and see if the RFC's have any opinion on whether its OK to accept illegal characters: security should certainly be the first priority.

clayrat:
So, regarding RFC and specs, I found these:
http://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2
http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
http://www.ietf.org/rfc/rfc2047.txt

clayrat:
The most recent one seems to be https://tools.ietf.org/html/rfc5987

bryce-anderson:
Page 5 of 5987 suggests we can drop/replace/pretty much whatever we want regarding illegal characters.

bryce-anderson:
Although I think that might be for decoding something like `%00`...

bryce-anderson:
I think this is ready to go if someone would please give it a review. For others, the TLDR is we will replace invalid characters with the replacement character '0xfffd'.

rossabaker:
:+1:, but I'll leave time in case @clayrat would like to weigh in.

refried:
Seems scary at a glance but I guess replacing with a character called “replacement character” is reasonable. :)

bryce-anderson:
I'm merging this in the evening (~5pm EST) unless there are objections.

clayrat:
Looks good :)

PR #38, opened 2016-03-15
Add InetSocketAddress to server chsocket address to server channel.

rossabaker:
This is so we can expose the bound port on the `BlazeServer`, for when it's constructed to listen on port 0.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
@@ -2,6 +2,7 @@ package org.http4s.blaze.channel
 
 
 import java.io.Closeable
+import java.net.InetSocketAddress
 import java.nio.channels.ClosedChannelException
 import java.util.concurrent.CountDownLatch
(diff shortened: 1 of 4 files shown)


rossabaker:
I could be shamed into pattern matching and looking for sensible defaults on these `asInstanceOf` calls.  I'm not happy about what happened here.  Has anybody ever seen a `SocketAddress` that is not an `InetSocketAddress` in the wild?

bryce-anderson:
The java type hierarchy only has one descendant from `SocketAddress`: `InetSocketAddress` so this seems fine. I imagine it was for forward compatibility reasons or maybe even custom sockets? At any rate, I think this is fine for now but am willing to revert it should we find a use case.

rossabaker:
It reminds me of `ServletRequest` vs. `HttpServletRequest`: an interface without a purpose.

rossabaker:
Less flippantly, `InetSocketAddress` is for IP.  Like `ServletRequest` is only ever implemented for HTTP, it seems that `SocketAddress` is only ever implemented for IP.

PR #41, opened 2016-06-17
Add streaming body support to the blaze-http model

bryce-anderson:
Numerous changes to the models used by the http project. The core point is the
movement to a reshape bodies that allows for streaming. Previously the entire
body was buffered which can result in wasted memory or requiring that bodies
be limited to a certain size which is not a very general solution. Instead,
the body is now fetched asynchronously by way of `Future`s. The model borrows
heavily from that of the Haskell WAI library.

Changes were also made to the way responses are built. Because the entire
response is no longer available (the body may be streaming in) we now use a
continuation style interface. This defines
```scala
type HttpService = HttpRequest => Future[ResponseBuilder]

sealed trait ResponseBuilder

case class HttpResponse(action: RouteAction) extends ResponseBuilder
case class WSResponseBuilder(pipeline: LeafBuilder[WebSocketFrame]) extends ResponseBuilder

trait RouteAction {
  def handle[Writer <: BodyWriter](commit: (HttpResponsePrelude => Writer)): Future[Writer#Finished]
}
```
The type of `RouteAction` requires the `close()` method of the `BodyWriter`
to be invoked in order to get the `Writer#Finished` type, ensuring cleanup
of resources.

Several smaller changes are also present and a non-inclusive list follows:
* Make a org.http4s.blaze.http.client package
* Client model is completely different: the client takes a continuation which
  receives the response and computes a Future[A], cleaning up the connection on
  completion of the future.
* Start refactoring client to support different methods of managing connections.
* Rename some types to make things _hopefully_ more readable.
* Move some things around to reduce disorder and help in type location.
* Use an explicit ExecutionContext in route execution (HttpServerStage).
* Renamed BasicHttpStage to Http2ServerStage: its not so basic anymore.
* Improve http2 conformance.
* Message bodies get a named type. Making it an object allows us to put helper
  methods on it such as the ability to accumulate the whole body.


bryce-anderson:
I know @rossabaker took a look at this when it was ~80% baked, so if there are no objections in a few days, I'll merge it.

rossabaker:
Did this get squashed? I'm looking for the other 20%.

bryce-anderson:
@rossabaker: I'm sorry, I'm terrible. I did rebase it onto master which killed your comments and the changes. I can probably git reflog it if you like.

rossabaker:
Nah, I'll look again fresh.

PR #42, opened 2016-06-26
Clean up some of the concurrency tools

bryce-anderson:
Documentation was incorrect in the actor section. Note: the actors are not
actually used in the library and are just there because its small and might
be useful in the future.

The trampoline executor has been reworked. The old behavior was based on
a simple `ArrayDequeue` which was allocated and destroyed as the executor was
needed and then not needed etc. This is pretty ugly for situations that only
submit a single `Runnable`. The new executor now unrolls the first three
elements of the queue which should cover the vast majority of cases. When
the first three elements spill over, an `ArrayDequeue` is created. After the
queue is emptied, the `ArrayDequeue` is removed to avoid memory leaks.


rossabaker:
It seems sound.  Does this cause any measurable benefit, or is this a gut-feeling optimization?

bryce-anderson:
Its a gut feeling which was inspired by com.twitter.util code: they use a queue unrolling strategy with some differences such as behavior in the event of failure and never clearing the `ArrayDequeue`.

The correct behavior on failure is an open question in my mind, but I like the idea of minimizing memory leaks associated with thread local scope.

rossabaker:
Alright, works for me.

PR #45, opened 2016-09-24
Work around the java Socket "memory leak"

bryce-anderson:
Problem

The java NIO implementation leaks direct memory because of the
way that it caches temporary direct ByteBuffers. This can lead
to OOM like errors.

Solution

Ensure that the scratch buffer is a direct ByteBuffer, and copy
heap buffers into it before writing, essentially stepping around
the logic that the JDK uses to write heap buffers.

Details

- Add additional allocation methods to BufferTools to be more
explicit.

- Use direct ByteBuffers for the scratch buffers in in NIO1 and NIO2
because they will most often be talking directly to the socket and
would necessitate a copy.

- Make tools for copying and fast forwarding ByteBuffers

- Fix the memory leak.

- Be awesome.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
@@ -231,6 +231,7 @@ class NIO1SocketServerGroup(pool: SelectorLoopPool) extends ServerChannelGroup {
           scratch.clear()
           BufferTools.copyBuffers(buffers, scratch)
           scratch.flip()
+
           val size = scratch.remaining()
           ch.write(scratch)
           val written = size - scratch.remaining()
--- a/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
@@ -5,24 +5,24 @@ package nio1
 import java.io.IOException
 import java.nio.ByteBuffer
 import java.nio.channels._
-import java.net.{InetSocketAddress, SocketAddress}
+import java.net.InetSocketAddress
 import java.util.Date
-import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.atomic.AtomicReference
 
 import org.http4s.blaze.channel.ChannelHead._
-import org.http4s.blaze.channel.nio1.NIO1HeadStage.{WriteError, Incomplete, Complete, WriteResult}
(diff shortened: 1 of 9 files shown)


bryce-anderson:
cc @hvesalai

rossabaker:
Is this fundamental enough to benefit from some benchmarks before and after?

bryce-anderson:
It looks like it took a little hit, I'll continue to look into it.

hvesalai:
@bryce-anderson LGTM

bryce-anderson:
@hvesalai, great! I think I still have a little bit of tuning to look into. When running load on the same computer I get some strange behavior if the scratch buffer is less than the message size: essentially I get 1/10th the performance with almost no CPU usage. Turning on `TCP_NODELAY` fixes it, but I don't know if I should change that default. (Really, it should be configurable...)

Do you have the ability to test this out before I merge it?

hvesalai:
Yeah, TCP_NODELAY should definitely be configurable. Did you try the latest patch (i.e. the writeLoop). I'm wondering how TCP_NODELAY should have such an impact on performance. Is the network layer lying about the buffer size (i.e. waiting for more data even though it signaled the buffer is already full).

Or is Java lying about what when the network buffers have been saturated?

bryce-anderson:
Note that the results I'm discussing are when running both load and server on the same host, so there is a good chance that actual TCP has nothing to do with it. Unfortunately I don't have access to a good wired network anymore and even the "bad performing" result was 500 MB/sec throughput.

hvesalai:
Could it be a test setup thing rather than actual difference in performance?

bryce-anderson:
Yes, I actually think its likely that its a test setup thing, but I don't currently have a way to test that hypothesis. :smile:

bryce-anderson:
Benchmark results:
```
         /ping                      /bigstring
Before   193750.96 (25.5 MB/sec)    1125.51 (121MB/sec)
After    195989.97                  1127.73

Req/sec. Separate Amazon m4.2xlarge instances for server and for load. wrk, 500 connections, 8 threads.
```
I don't think this change makes any significant performance difference.

rossabaker:
I am underqualified on NIO, but it seems reasonable to me.

bryce-anderson:
An important note is that none of the results above were saturating the CPU. I'll leave this for a little bit to give @hvesalai a chance to object, but plan on merging tonight or tomorrow.

bryce-anderson:
I've re-targeted this against 0.12.1-release (see #47). I also dropped some of helpers I added for constructing `ByteBuffer`s. Might be worth another look.

hvesalai:
:+1: on this patch and also the new benchmark results :-)

PR #47, opened 2016-09-26
Work around the java Socket "memory leak"

bryce-anderson:
Problem

The java NIO implementation leaks direct memory because of the
way that it caches temporary direct ByteBuffers. This can lead
to OOM like errors.

Solution

Ensure that the scratch buffer is a direct ByteBuffer, and copy
heap buffers into it before writing, essentially stepping around
the logic that the JDK uses to write heap buffers.

Details

- Add additional allocation methods to BufferTools to be more
explicit.

- Use direct ByteBuffers for the scratch buffers in in NIO1 and NIO2
because they will most often be talking directly to the socket and
would necessitate a copy.

- Make tools for copying and fast forwarding ByteBuffers

- Fix the memory leak.

- Add a write loop so we saturate the tcp buffer without doing
excessive copies that are then discarded.

- Be awesome.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
@@ -5,24 +5,24 @@ package nio1
 import java.io.IOException
 import java.nio.ByteBuffer
 import java.nio.channels._
-import java.net.{InetSocketAddress, SocketAddress}
+import java.net.InetSocketAddress
 import java.util.Date
-import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.atomic.AtomicReference
 
 import org.http4s.blaze.channel.ChannelHead._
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
```diff
@@ -125,7 +125,7 @@ class NIO1SocketServerGroup(pool: SelectorLoopPool) extends ServerChannelGroup {
 
               // check to see if we want to keep this connection
               if (acceptConnection(address)) {
-                clientChannel.setOption[java.lang.Boolean](java.net.StandardSocketOptions.TCP_NODELAY, false)
+                clientChannel.setOption[java.lang.Boolean](java.net.StandardSocketOptions.TCP_NODELAY, true)
```
hvesalai: I think `true` here is the best default since Blaze does try to saturate the TCP/IP buffer if there is enough data. If not, then then there is no point for the network layer to wait for it either.

hvesalai: s/buffer/MTU size/

PR #51, opened 2016-10-30
Explicitly set targets in Java and Scala

rossabaker:
We used to support Java 1.7 on Scala 2.10, but never applied it to Java.
De facto, we already released for Java 1.8.  Making it explicit so this
doesn't kill us in 1.9.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -48,7 +48,12 @@ lazy val examples = Project("blaze-examples",
 val dontPublish = packagedArtifacts := Map.empty
 
 val jvmTarget = TaskKey[String]("jvm-target-version", "Defines the target JVM version for object files.")
-jvmTarget in ThisBuild := "1.8"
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -48,12 +48,7 @@ lazy val examples = Project("blaze-examples",
 val dontPublish = packagedArtifacts := Map.empty
 
 val jvmTarget = TaskKey[String]("jvm-target-version", "Defines the target JVM version for object files.")
-jvmTarget in ThisBuild := {
-  VersionNumber(scalaVersion.value).numbers match {
-    case Seq(2, 10, _*) => "1.7"
-    case _ => "1.8"
-  }
-}
+jvmTarget in ThisBuild := "1.8"
```
bryce-anderson: Looks like this was actually necessary because 2.10 doesn't want to emit java 8 bytecode.

PR #58, opened 2017-01-15
Http refinement

bryce-anderson:
I think this is better than what is currently in master, but I don't think the blaze http story is 100% baked. Things that will likely change:

- The `HttpServerStage` is less complex mostly because it all got shoveled into `HttpServerCodec`. I may experiment with breaking down the byte stream into structured messages (Request, Body, Response, etc) which will be more reusable, but may not due too much to lower complexity.
- The writers are part of what makes `HttpServerCodec` seem so cluttered, so they may be split out into their own file. Before that happens a better model around the locking would be smart.


http/src/main/java/org/http4s/blaze/http/parser/Http1ServerParser.java
```diff
@@ -48,6 +48,10 @@ public final boolean requestLineComplete() {
         return _lineState == LineState.END;
     }
 
+    public final boolean isReset() {
```
rossabaker: I don't think I'd think something `isReset` if it's never been used.  `isStart`, maybe?

http/src/main/scala/org/http4s/blaze/http/HttpServerStage.scala
```diff
@@ -1,262 +1,113 @@
...
-    val sb = new StringBuilder(512)
-    sb.append("HTTP/").append(1).append('.')
-      .append(minor).append(' ').append(400)
-      .append(" Bad Request\r\n\r\n")
-
-    channelWrite(ByteBuffer.wrap(sb.result().getBytes(StandardCharsets.ISO_8859_1)))
-      .onComplete(_ => shutdownWithCommand(Cmd.Disconnect))
+  private def make5xx(error: Throwable): RouteAction = {
+    logger.error(error)("Failed to service request. Sending 500 response.")
+    RouteAction.String(500, "Internal Server Error", Nil, "Internal Server Error.")
```
rossabaker: I hope we're not making enough errors for it to matter, but this could be a shared 500.

http/src/main/scala/org/http4s/blaze/http/HttpServerStage.scala
```diff
@@ -1,262 +1,113 @@
...
-          false
-        }
-
-      case None if minor == 0 => false
-      case None               => true
+  // Determine if this request requires the connection be closed
+  private def requestRequiresClose(request: HttpRequest): Boolean = {
+    val connHeader = request.headers.find { case (k, _) => k.equalsIgnoreCase(HeaderNames.Connection) }
+    if (request.minorVersion == 0) connHeader match {
+      case Some((_,v)) => !v.equalsIgnoreCase("keep-alive")
```
rossabaker: Corner case, but do these values need to be trimmed?

http/src/main/scala/org/http4s/blaze/http/RouteAction.scala
```diff
@@ -26,53 +28,86 @@ object RouteAction {
     *             before requesting another chunk. Termination is signaled by an __empty__ `ByteBuffer` as
     *             defined by the return value of `ByteBuffer.hasRemaining()`.
     */
-  def streaming(code: Int, status: String, headers: Headers)(body: () => Future[ByteBuffer]): HttpResponse = HttpResponse(
+  def Streaming(code: Int, status: String, headers: Headers)
```
rossabaker: It's weird to me that these are capitalized.

http/src/main/java/org/http4s/blaze/http/parser/Http1ServerParser.java
bryce-anderson: You're right, `isStart` feels better.

http/src/main/scala/org/http4s/blaze/http/HttpServerStage.scala
bryce-anderson: I'm not sure if you mean that it could be added to the `RouteAction` or not, but that seems like a good idea.

http/src/main/scala/org/http4s/blaze/http/HttpServerStage.scala
bryce-anderson: They should have already been trimmed by the parser.

http/src/main/scala/org/http4s/blaze/http/RouteAction.scala
bryce-anderson: Yeah, I agree. Do you have a better idea in mind? I didn't get inspired and to some extent don't care: hopefully very few to no folks are using these for anything other than tinkering. 

My hope is for this to be solid and just complete enough to serve as a foundation for http4s proper so we don't have as much low level stuff laying around in there.

bryce-anderson:
@rossabaker, do you think there is anything holding this back from hitting master? I don't think the story is complete, but I think its better than the status quo. I'm also interested in other feedback. I haven't addressed your issue with the name casing, but am happy to make them lowercase if you feel strongly about it.

build.sbt
```diff
@@ -38,13 +38,14 @@ lazy val examples = Project("blaze-examples",
...
       libraryDependencies += alpn_boot,
+      fork := true, // necessary to add ALPN classes to boot classpath
 
       // Adds ALPN to the boot classpath for Http2 support
       javaOptions in run <++= (managedClasspath in Runtime) map { attList =>
         for {
           file <- attList.map(_.data)
           path = file.getAbsolutePath if path.contains("jetty.alpn")
-        } yield { println(path); "-Xbootclasspath/p:" + path}
+        } yield { println(s"Alpn patth: $path"); "-Xbootclasspath/p:" + path}
```
rossabaker: s/patth/path/

http/src/main/scala/org/http4s/blaze/http/Http1ServerCodec.scala
```diff
@@ -295,7 +298,9 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
       private var prelude: StringBuilder,
       maxCacheSize: Int) extends InternalWriter {
 
-    prelude.append("transfer-encoding: chunked\r\n\r\n")
+    // The transfer-encoding header will be the last header, and before each write,
+    // we will append a '\r\n{length}\r\n
+    prelude.append("transfer-encoding: chunked\r\n")
```
rossabaker: Is there any reason we print headers in lower case instead of their more canonical form?  It doesn't really matter, but I find it curious.

bryce-anderson: Http2 requires the be transmitted in lower case, so I've been trying to do that consistently. Do you think it's worth changing?

rossabaker: Maybe.  It's a quirk that could help with server fingerprinting, though a dedicated fingerprinter is likely to stay one step ahead of our efforts to blend into the crowd.

If it's simple and doesn't break HTTP/2, yes.  If not, nah.

bryce-anderson: A number of headers, including `content-length`, are defined in a shared location and are used by http/2, so I'm going to leave it as is so that the case of the `content-length` header doesn't change depending on whether its added automatically by the writer or by another mechanism.

PR #61, opened 2017-01-27
Add ability to configure channel options

### LINKED ISSUE #46: Add ability to configure socket options
Being able to tuning them would be nice.

bryce-anderson:
Fixes #46

Add `ChannelOptions`, a collection of ChannelOption->value pairs
that can be applied to a socket connection. These are passed to
the common vectors of socket generation to configure the socket
after its connected.

Also perform some minor API cleanup: make classes final, etc.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -111,6 +111,7 @@ lazy val mimaSettings = Seq(
 /* dependencies */
 lazy val dependencies = Seq(
   libraryDependencies += specs2 % "test",
+  libraryDependencies += specs2Mock % "test",
   libraryDependencies += logbackClassic % "test",
   libraryDependencies += log4s
 )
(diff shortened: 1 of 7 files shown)


bryce-anderson:
I'm going to merge this one tomorrow or the next day barring objections: I don't think this one is contentious, and its definitely better than the situation we have now.

bryce-anderson:
Cool, merging.

@rossabaker: one thing to note is that nio2 now defaults to disable nagle's algorithm. This should primarily effect the client which uses nio2 exclusively. I generally think its a nuisance, but if you notice any degraded performance that can be traced back to sending a ton of small TCP packets, this could be it. On the other hand, it might speed the client up for GET requests since they are generally small and might be unnecessarily delayed by the algorithm.

PR #64, opened 2017-01-30
Add method `takeSlice` to BufferTools

bryce-anderson:
`takeSlice(ByteBuffer, Int)` makes it easier to deal
with framed content by allow one to easily slice away
a sub-buffer.

This exposed a problem encoding header continuation
frames: an oversized HEADERS frame was not split correctly,
so it was fixed and some tests added.


bryce-anderson:
Wow, I forgot about this one. It turns out that it has already made it into master: https://github.com/http4s/blaze/blob/master/core/src/main/scala/org/http4s/blaze/util/BufferTools.scala#L90

PR #67, opened 2017-03-06
Improvements to core from http2 improvements

bryce-anderson:
Some improvements that helped along the way in making the `new-client-http2`.


bryce-anderson:
Considering this should be pretty non-controversial, I plan to merge this in a few days barring objects.

core/src/main/scala/org/http4s/blaze/pipeline/stages/OneMessageStage.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import org.http4s.blaze.pipeline.MidStage
+
+import scala.concurrent.Future
+
+/** Holds a single element that when read, will eject itself
+  * from the pipeline.
+  * @note There is an intrinsic race condition between this stage
+  *       removing itself and write commands, so this stage should
+  *       only be used under half duplex conditions.
```
rossabaker: Would it be possible to enforce that with a `Nothing` parameter?  Probably no way to make the variance work out on both sides...

bryce-anderson: I'm sorry, I don't understand. Enforce the half duplex requirement? Maybe I used the wrong terminology, but I meant that they shouldn't be used concurrently. Maybe I should rephrase it.

ChristopherDavenport: I agree with a slight rephrase, it's a bit hard to follow from the comment to the intent you just expressed.

PR #68, opened 2017-03-06
WIP: HTTP(1/2) Client and Server Refactor

bryce-anderson:
## WIP: do not merge.

We are in need of a better HTTP2 implementation and the HTTP/1.x
client is a bit lacking. This branch represents an incomplete step
forward in regard to these two goals.

### What the PR Accomplishes
We have a working client and server for both HTTP/1.x and HTTP/2.

### Vague List of Things Left to Do
- There a numerous stubs still laying around, mostly in HTTP2 code. Finish them.
- Very little of the new code is tested. Add some tests.

### More Concrete Steps

Start pulling the code into a piece at a time. This will require tests.

we need to decide if these are the interfaces we want. Most of of the implementation
details are private but the interfaces will be what is used by http4s proper. Specific
interfaces of note:
- The `HttpClientSession` family of interfaces
- Do we need to provide a session manager in blaze, or is that a http4s problem?
- Websockets. Do we need them here at the blaze level? Do we have the hooks
  to build them if blaze proper is uninterested in handling them? (no...)


rossabaker:
I need to spend a lot more time on this, but some quick hits:

* http4s' `PoolManager` returns and invalidates connections in a Task.  I'm a little surprised they're not `Future[Unit]` here.
* The http4s session manager is going to need to be in Task.  If it's in blaze, it's easier for http4s to wrap it.  If it's in http4s, it doesn't need a wrapper.  I'm not sure which side I'm arguing.
* I hate that the current web socket implementation has frames based on mutable arrays, which makes me want to move more of that to http4s.  But being able to do it is certainly important.

bryce-anderson:
I think the differences regarding returning and invalidating sessions has to do with the fundamental difference between `Task` and `Future`: the execution of a `Task` is the users responsibility while if you get back a `Future`, its up to the library to schedule the execution. Therefore, the only thing that `Future[Unit]` gives you over just a straight `Unit` is the ability to defer the work internally without blocking. I didn't foresee that being significant enough to warrant the overhead imposed by bouncing through a `Future`, but maybe I'm being shortsighted.

rossabaker:
I was thinking about being asynchronously notified when the work is complete.  If it's a blocking operation under the hood, it doesn't buy much -- the `Task` wrapper can do the scheduling then.

bryce-anderson:
If we're talking about
```scala
PoolManager {
  def release(connection: A) = Task[Unit] = ???
  def invalidate(connection: A): Task[Unit] = ???
   ...
}
```
I see little value to being notified when those operations complete, and actually think that its none of the users business.

rossabaker:
I spent too much time in Akka cursing `Unit` when I needed an ack back that I treat them all with suspicion now.  The best use case I can think of is connection metrics, but I think that will be handled by other means anyway.  I'm willing to try it this way.

ChristopherDavenport:
It's not about notification to the user it's about the control of side-effects when those effects are streamed. Streams can take tasks and execute their side effects transparently in a known fashion. Where futures eager evaluation leaves no control of execution, except the pool/context on which it is executed.

The implementation I have in cats utilized without the `Task` in `Connection`, so that we could get ourselves operational, however I was reticent to put it in place. Now in blaze-client it is simple to wrap and control anything you have down in blaze through fs2, since there are conversions(although one direction is unsafe, to/from `Task`). 

It's not about user level notification and control but the controlled execution of effects within our application code.

http/src/main/scala/org/http4s/blaze/http/http2/HeaderAggregatingFrameHandler.scala
```diff
@@ -0,0 +1,173 @@
...
+                                    priority: Option[Priority],
+                                    endHeaders: Boolean,
+                                    endStream: Boolean,
+                                    buffer: ByteBuffer): Http2Result = {
+
+    if (inHeaderSequence()) {
+      return PROTOCOL_ERROR.goaway(s"Received HEADERS frame while in in headers sequence. Stream $streamId").toError
+    }
+
+    if (buffer.remaining() > inboundSettings.maxHeaderListSize) {
```
bryce-anderson: TODO: the `maxHeaderListSize` is for uncompressed headers. This should always be smaller than or at least approximately the same size as the uncompressed size, so maybe its ok?

bryce-anderson:
@mehmetgunturkun, this is the branch where I'm working on H2 support in blaze, both client and server. What is left to do is predominantly writing tests and addressing TODO's. I think we need a good test harness to perform some end-to-end tests with the new h2 client and server and a lot of stuff is uncovered by unit tests.

bryce-anderson:
I'm going to start breaking this up and landing chunks at a time to make it easier to review. I'll leave this one open for reference.

bryce-anderson:
Squashed and rebased against 6621a0c251105ee0. Going to start breaking into PR's.

bryce-anderson:
After fixing the 2.10 build...

rossabaker:
Should I be looking at this again or waiting for splits?

bryce-anderson:
I'm putting up splits and merging this branch with master as they go in.
Hopefully another split will go up this evening.

PR #70, opened 2017-04-02
Add full text of the Apache-2 license

bryce-anderson:
This is the proper thing to do when adding the LICENSE file to the
root directory of a OSS project.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,15 +1,202 @@
-This software is licensed under the Apache 2 license, quoted below.
 
-Copyright (C) 2013-2014 http4s.org
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Licensed under the Apache License, Version 2.0 (the "License"); you may not
-use this file except in compliance with the License. You may obtain a copy of
-the License at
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,15 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
ChristopherDavenport: I'm not versed on license application. Is this supposed to remain blank or be filled out with project details.

bryce-anderson: Its not in the the few examples I've checked which includes the twitter/finagle project, and http4s/http4s. I believe it is supposed to be an example of how to add the license info to your source files.

ChristopherDavenport: If it works there, it should work here.

Issue #71, opened 2017-04-28
Blaze client stack overflow for user generated SSL cert


TimothyKlim:
I'm trying to connect through http4s-blaze-client (0.17.0) to my kubernetes https api with user generated ssl cert and I'm getting stack overflow 😢 : https://gist.github.com/TimothyKlim/0b05e9ab46cd63a2c567a97bb6a026b1

bryce-anderson:
Do you know how large this cert is? Its definitely a SOE, but I'm concerned that there is a deeper underlying problem and trampolining it would turn it into a infinite loop.

bryce-anderson:
@TimothyKlim If you could provide an example for reproduction, that would help figure out if we just need to break the recursion or fix a larger problem.

bryce-anderson:
@TimothyKlim, https://github.com/http4s/blaze/pull/72 should fix it if its truly just a SOE. Is there a chance you can test it?

bryce-anderson:
Alright, merged and released in 0.12.5 which is on its way to Central. @TimothyKlim, If you want to follow up with the http4s proper project they should include it in releases shortly. In the meantime, you can include blaze-core 0.12.5 manually.
Thanks for the bug report and promptly checking the patch!

PR #72, opened 2017-04-29
Trampoline SSLStage handshake

bryce-anderson:
Fixes issue #71

To prevent SOEs, we make to looping parts of `SSLStage.sslHandshake`
tail recursive and wrap that in the try/catch block.

If this solves the issue, it should also be merged into master.


TimothyKlim:
Tested it with local snapshot and it is works! Thanks!

bryce-anderson:
Sweet. I'll cherry pick this into master as well and get a release rolling.

PR #73, opened 2017-05-12
Version updates

ChristopherDavenport:
Went to do a simple update of a few of the dependencies, ended up needing to do a bit of a refactor on the build in order to get it all working as the deprecation window had expired. 

Everything seems to compile nicely now on my machine. However when build.sbt get changed I'd love as many eyes as possible.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,9 +66,9 @@ organization in ThisBuild := "org.http4s"
 
 version in ThisBuild := "0.14.0-SNAPSHOT"
 
-scalaVersion in ThisBuild := "2.11.8"
+scalaVersion in ThisBuild := "2.11.11"
 
-crossScalaVersions in ThisBuild := Seq("2.10.6", scalaVersion.value, "2.12.0")
+crossScalaVersions in ThisBuild := Seq("2.10.6", scalaVersion.value, "2.12.2")
 
 description in ThisBuild := "NIO Framework for Scala"
 
--- a/build.sbt
+++ b/build.sbt
@@ -1,111 +1,43 @@
 import scala.util.Properties
 
-/* Projects */
-lazy val blaze = project
-  .in(file("."))
-  .settings(dontPublish:_*)
-  .aggregate(core, http, examples)
+/* Global Build Settings */
 
-lazy val core = Project("blaze-core",
-  file("core"),
-  settings = publishing ++ mimaSettings ++ dependencies)
-
-lazy val http = Project("blaze-http",
-  file("http"),
-  settings = publishing ++ mimaSettings ++ dependencies ++ Seq(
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -117,33 +49,33 @@ lazy val dependencies = Seq(
...
+lazy val asyncHttpClient     = "org.asynchttpclient"        %  "async-http-client"   % "2.0.32"
 
 
 // Needed for Http2 support until implemented in the JDK
 lazy val alpn_api            = "org.eclipse.jetty.alpn"     % "alpn-api"             % "1.1.3.v20160715"
 
 // Note that the alpn_boot version is JVM version specific. Check the docs if getting weird errors.
 // Also note that only java8 and above has the require cipher suite for http2.
-lazy val alpn_boot           = "org.mortbay.jetty.alpn"     % "alpn-boot"            % "8.1.9.v20160720"
+lazy val alpn_boot           = "org.mortbay.jetty.alpn"     % "alpn-boot"            % "8.1.11.v20170118"
```
bryce-anderson: Not a problem but just in case you weren't aware, this version is JVM version dependant.

ChristopherDavenport:
Not sure why it is failing, maybe something to do with being a PR?

Here was my [build on the same commit](https://travis-ci.org/ChristopherDavenport/blaze/builds/231735419)

build.sbt
rossabaker: How JVM-dependent?  "Requires Java 8+"?  "Requires Java 8.* but no Java 9"?  "Requires this particular patch of the JVM?"

rossabaker:
The build failure is https://github.com/travis-ci/travis-ci/issues/7703, I think.

bryce-anderson:
Looks like @rossabaker is right on the money. If this builds locally, I'm good to go. On the other hand, doesn't seem like a rush so your call @ChristopherDavenport.

ChristopherDavenport:
This publishes fine locally. Checked before pushing. 

Easy solution is I can push the fix for the travis build so we won't have any troubles. When it fixes we can always roll it out since CI wont have an effect on the artifacts.

build.sbt
bryce-anderson: Every JVM version has a corresponding alpn_boot version because this package does some hackery to get itself integrated with the Java SSL library: https://www.eclipse.org/jetty/documentation/9.4.x/alpn-chapter.html#alpn-versions

bryce-anderson:
@ChristopherDavenport I'm 👍 on this. If you like, it might be nice to add a comment next to the permissions change in the travis file so its not a mystery to folks in the future.

ChristopherDavenport:
I'll add a reference comment so people know why it's there. Always a good plan.

PR #79, opened 2017-06-07
Use SbtRig for Deployments

ChristopherDavenport:
Came with a lot of new scalac options that were updated for as well. 

Implicit Widening
Procedure Syntax
Classes/Traits in Package Objects
Warn Value Discard 

Were the ones that had significant changes as a result.  Would like @rossabaker to make sure the release command looks right and that everything is in place to run a deploy. 

Can still add Coursier if we would like to make downloading more expedient, though this was quite a PR by itself.


core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -277,33 +281,36 @@ final class SSLStage(engine: SSLEngine, maxWrite: Int = 1024*1024) extends MidSt
...
+        val out = new ArrayBuffer[ByteBuffer]
+        writeLoop(data, out) match {
+          case SSLSuccess if BufferTools.checkEmpty(data) =>
+            p.completeWith(channelWrite(out))
+
+          case SSLSuccess => // must have more data to write
             channelWrite(out).onComplete {
-              case Success(_) => sslHandshake(readData, r)
-              case Failure(t) => handshakeFailure(t)
+              case Success(_) => continueWrite(data, p)
```
ChristopherDavenport: Is this a reversion that needs to be fixed, I was working with older data and the rebase may have overridden.

http/src/main/scala/org/http4s/blaze/http/BlazeServerParser.scala
```diff
@@ -10,12 +10,12 @@ import scala.collection.immutable.VectorBuilder
 
 private[blaze] object BlazeServerParser {
 
-  final case class Prelude[Headers](
+  final case class Prelude[HeadersT](
```
ChristopherDavenport: Fixes conflicting type names in same Scope

.travis.yml
```diff
@@ -1,23 +1,40 @@
 language: scala
+
 scala:
+  - 2.12.2-bin-typelevel-4
+  - 2.11.11-bin-typelevel-4
```
bryce-anderson: What is the purpose of compiling 2.12 and 2.11 with typelevelc while we still need to build 2.10 with lightbend?

core/src/main/scala/org/http4s/blaze/pipeline/stages/monitors/ConnectionMonitor.scala
```diff
@@ -48,17 +48,17 @@ abstract class ConnectionMonitor {
     }
 
     def writeRequest(data: ByteBuffer): Future[Unit] = {
-      bytesOutBound(data.remaining)
+      bytesOutBound(data.remaining.toLong)
```
bryce-anderson: Just for my own information, is auto widening of ints/floats a warning or or error with lightbendc?

core/src/test/scala/org/http4s/blaze/util/ExecutionSpec.scala
```diff
@@ -48,6 +48,7 @@ class ExecutionSpec extends Specification {
       i must be equalTo(10)
     }
 
+    /* This Job Can't Fail As the i+=1 is Dead Code
```
bryce-anderson: Would you like to fix it instead of just commenting it out? I think the goal was to be able to submit a job that is going to fail anyway, so removing the counter and just ending the test with `ok` is probably fine.

http/src/main/scala/org/http4s/blaze/http/client/package.scala
```diff
@@ -1,54 +1,3 @@
...
-    private def getCharset(hs: Headers): String = {
-      headers.collectFirst {
-        case (k, v) if k.equalsIgnoreCase("content-type") => charsetRegex.findFirstIn(v)
-      }.flatten.getOrElse("UTF-8")
-    }
-  }
-
-  private val charsetRegex = "(?<=charset=)[^;]*".r
-}
+package object client {}
```
bryce-anderson: I'm not proud that `ClientResponse` was in a package object. Do we even need the package object, or could we just delete this file?

core/src/main/scala/org/http4s/blaze/pipeline/stages/monitors/ConnectionMonitor.scala
rossabaker: Yes. It's `-Ywarn-numeric-widen`.

.travis.yml
rossabaker: I can't think of anywhere that blaze benefits featurewise.  It does make it easier to eventually spin off an http4s organization plugin to make repos consistent.  This becomes more compelling if we go polyrepo with the support modules in http4s.  Also, the more projects build with typelevelc, the more we're helping to test it.

I can argue both ways flimsily, but am likely to side with whoever pounds on the table the loudest.

http/src/main/scala/org/http4s/blaze/http/client/package.scala
ChristopherDavenport: Fairly certain we can delete.

core/src/test/scala/org/http4s/blaze/util/ExecutionSpec.scala
ChristopherDavenport: Will grab this asap.

.travis.yml
ChristopherDavenport: I don't care either way honestly. I opted for consistency alone.

ChristopherDavenport:
Seriously hate that 2.10.6 difference in Errors on Return Value. 

Have to remember to test the specific version or else stuff slips by.

bryce-anderson:
Is this ready to land? I'm going to start landing https://github.com/http4s/blaze/pull/68 and there will be some serious merge conflicts, so if this is ready to go we should just get it done to make the conflicts in just one spot.

ChristopherDavenport:
I believe so. Might be quite a few bumps and bruises with the process. Specifically the 2.10 branch. But I believe my last round of updates got this up to par.

rossabaker:
Note that there is a version 4 of rig now. Before that, codecov is completely broken.

bryce-anderson:
I'm not to worried about codecov, but if its simple to switch we should do it. @ChristopherDavenport, I'll let you drive the merging (hopefully a squash+merge) of this one. I'd like to start breaking http2 stuff up this weekend, but if not so be it.

PR #83, opened 2017-07-21
Changed 'to' to 'from' for clarity

### LINKED ISSUE #82: Confusing log message when accepting incoming connections
I'm using blaze via http4s and have found that whenever an incoming connection is accepted, messages like the following are logged:
```
[NIO1SocketServerGroup Acceptor] INFO  o.h.blaze.channel.ServerChannelGroup - Connection to /0:0:0:0:0:0:0:1:35540 accepted at Fri Jul 21 15:24:57 BST 2017.
```
This makes it sound like the server is making connections elsewhere, whereas changing the "to" to "from" would make what actually happened much clearer.

ChristopherDavenport:
Fixes #82


cquiroz:
Unrelated to this PR but why to use the old `Date` rather than the new `java.time` API?

bryce-anderson:
Ugg... Sorry I missed this in the inbox! Thanks for fixing it @ChristopherDavenport.

PR #85, opened 2017-07-30
Fix a race condition in ServerChannel

bryce-anderson:
When calling ServerChannel.join(), its not clear that tasks may not be
finished executing before the call returns because it's up to the caller
of `close()` to execute all the tasks.

We fix this by making `join()` wait until all tasks have been executed
before returning. To improve the user experience, we also allow passing
an `ExecutionContext` on which to run the shutdown hooks.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
@@ -1,67 +1,114 @@
 package org.http4s.blaze.channel
 
-
 import java.io.Closeable
 import java.net.InetSocketAddress
-import java.util.concurrent.CountDownLatch
-import java.util.concurrent.atomic.AtomicReference
-
-import scala.annotation.tailrec
-import scala.util.control.NonFatal
-
 import org.log4s.getLogger
+import org.http4s.blaze.util.Execution
+import scala.util.control.NonFatal
+import scala.concurrent.ExecutionContext
 
 /** Representation of a bound server */
 abstract class ServerChannel extends Closeable { self =>
+  import ServerChannel.State._
+
   protected val logger = getLogger
 
-  private val shutdownHooks = new AtomicReference[Vector[()=> Unit]](Vector.empty)
+  private var state: State = Open
+
+  private val shutdownHooks = new scala.collection.mutable.Queue[Hook]
 
   /** Close out any resources associated with the [[ServerChannel]] */
   protected def closeChannel(): Unit
 
-  /** Close the [[ServerChannel]] and execute any shutdown hooks */
+  /** The bound socket address for this [[ServerChannel]] */
+  def socketAddress: InetSocketAddress
+
+  /** Close the [[ServerChannel]] and execute any shutdown hooks
+    *
+    * @note this method is idempotent
+    */
   final def close(): Unit = {
-    closeChannel()
-    runShutdownHooks()
+    val run = shutdownHooks.synchronized {
+      if (state != Open) false
+      else {
+        state = Closing
+        // Need to close the channel, get all the hooks, then notify the listeners
+        closeChannel()
+        true
+      }
+    }
+
+    // We move this outside of the lock to ensure that we don't schedule work
+    // inside the lock since the EC could run the task immediately.
+    if (run) scheduleHook()
   }
 
-  /** Wait for this server channel to close */
-  final def join(): Unit = {
-    val l = new CountDownLatch(1)
-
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
```diff
@@ -1,67 +1,114 @@
...
+        })
       }
     }
   }
+}
 
-  /* Return the bound socket address for this server channel */
-  def socketAddress: InetSocketAddress
+private object ServerChannel {
+  object State extends Enumeration {
```
rossabaker: Wow, you don't see these very often.  Any reason for this over a sealed trait?

ChristopherDavenport: :arrow_up:

core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
```diff
@@ -1,67 +1,114 @@
...
-    if (hooks != null) {
-      hooks.foreach { f =>
-        try f()
-        catch { case NonFatal(t) => logger.error(t)(s"Exception occurred during Channel shutdown.") }
+  private case class Hook(task: () => Unit, ec: ExecutionContext)
+
+  // Bundles the task of executing a hook and scheduling the next hook
+  private[this] class HookRunnable(hook: () => Unit) extends Runnable {
+    override def run(): Unit = {
+      try hook()
```
ChristopherDavenport: What guarantees to we have about `hook` here. 

If a hook is created in such a way it is put back in queue, this could create an infinite loop. As The `HookRunnable` is run outside of the synchronized lock for obvious lock contention reasons.

bryce-anderson: Thats not true: hooks are only run after the state has transitioned to closing, after which hooks are no longer added to the shutdown tasks (line 62).

Regardless, nothing stops a user from making an infinite loop in a hook:
```scala
.addShutdownHook { () => while(true) { /* LOL */ } }
```

core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
bryce-anderson: To avoid making 4 new classes/traits to clutter the classpath, which I suppose is a drop in the ocean for Scala. I'd definitely choose the sealed + case objects pattern of I did any matching, but it's all equality checks.
Is this a blocker for either of you? I'm not strongly opinionated about it, this just felt 'right' and is less code (by 1 line).

core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
```diff
@@ -1,67 +1,114 @@
...
-    closeChannel()
-    runShutdownHooks()
-  }
+  /** The bound socket address for this [[ServerChannel]] */
+  def socketAddress: InetSocketAddress
 
-  /** Wait for this server channel to close */
-  final def join(): Unit = {
-    val l = new CountDownLatch(1)
+  /** Close the [[ServerChannel]] and execute any shutdown hooks
```
bryce-anderson: Should change this to 'schedule' any shutdown hooks.

core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
rossabaker: Not a blocker for me, but I think these are on the way out in Scala.  I bet this gets rewritten when we port to 2.14 or Dotty or whatever. :)

rossabaker:
Oh, I see how it works now.  I thought I saw these hooks running concurrently, but we don't kick off the next until we complete the previous, and then we close on the final scheduling.

bryce-anderson:
It's not a terrible idea to launch them concurrently (if the EC's supports that). TBH, I didn't even think about it. WDYT?

core/src/main/scala/org/http4s/blaze/channel/ServerChannel.scala
ChristopherDavenport: Definitely not a blocker.

bryce-anderson: "On the way out" is good enough for me. Changed.

bryce-anderson:
Feedback addressed (I think) so I'll merge this in the morning barring objections.

PR #86, opened 2017-07-30
Add BodyReader and ALPNTokens

bryce-anderson:
BodyReader is going to be the core abstraction for extracting data from
a HTTP message body and ALPNTokens are going to be necessary for
performing ALPN negotiation between HTTP/1.x and HTTP/2.

### DIFF
--- a/http/src/main/java/org/http4s/blaze/http/parser/BodyAndHeaderParser.java
+++ b/http/src/main/java/org/http4s/blaze/http/parser/BodyAndHeaderParser.java
@@ -87,6 +87,7 @@ public abstract class BodyAndHeaderParser extends ParserBase {
         return _hstate == HeaderState.END;
     }
 
+    /** Determine if the parser is in a state that is guarenteed to not produce any more body content */
     public final boolean contentComplete() {
         return mustNotHaveBody() ||
             _endOfContent == EndOfContent.END ||
@@ -386,6 +387,7 @@ public abstract class BodyAndHeaderParser extends ParserBase {
                     final int buff_size = in.remaining();
 
                     if (remaining_chunk_size <= buff_size) {
+                        // End of this chunk
                         ByteBuffer result = submitPartialBuffer(in, (int)remaining_chunk_size);
                         _contentPosition = _chunkLength = 0;
                         _chunkState = ChunkState.CHUNK_LF;
@@ -426,6 +428,7 @@ public abstract class BodyAndHeaderParser extends ParserBase {
 
     /** Manages the buffer position while submitting the content -------- */
 
+    // TODO: do we care about these being `read-only`? It does ensure any underlying array is hidden...
     private ByteBuffer submitBuffer(ByteBuffer in) {
         ByteBuffer out = in.asReadOnlyBuffer();
         in.position(in.limit());
@@ -436,18 +439,9 @@ public abstract class BodyAndHeaderParser extends ParserBase {
         // Perhaps we are just right? Might be common.
         if (size == in.remaining()) {
             return submitBuffer(in);
+        } else {
+            return BufferTools.takeSlice(in, size).asReadOnlyBuffer();
         }
-
-        final int old_lim = in.limit();
-        final int end = in.position() + size;
-
-        // Make a slice buffer and return its read only image
-        in.limit(end);
-        ByteBuffer b = in.slice().asReadOnlyBuffer();
-        // fast forward our view of the data
-        in.limit(old_lim);
-        in.position(end);
-        return b;
     }
 
 }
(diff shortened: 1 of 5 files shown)


http/src/main/java/org/http4s/blaze/http/parser/BodyAndHeaderParser.java
```diff
@@ -87,6 +87,7 @@ public final boolean headersComplete() {
         return _hstate == HeaderState.END;
     }
 
+    /** Determine if the parser is in a state that is guarenteed to not produce any more body content */
```
rossabaker: *guaranteed

http/src/main/scala/org/http4s/blaze/http/BodyReader.scala
```diff
@@ -0,0 +1,125 @@
...
+  /** Examine whether the `MessageBody` may yield additional data.
+    *
+    * This may be a result of being discarded, failure, or deletion of the data stream.
+    *
+    * Because `MessageBody` is async it is not, in general, possible to definitively determine
+    * if more data remains in the stream. Therefore, the contract of this method is that a return
+    * value of `true` guarantees that no more data can be obtained from this `MessageBody`, but a
+    * return value of `false` does not guarantee more data.
+    */
+  def isEmpty: Boolean
```
rossabaker: I'm not real thrilled with the name in light of that comment.  `isExhausted`, maybe?

http/src/main/scala/org/http4s/blaze/http/BodyReader.scala
```diff
@@ -0,0 +1,125 @@
...
+  val EmptyBodyReader: BodyReader = new BodyReader {
+    override def discard(): Unit = ()
+    override def apply(): Future[ByteBuffer] = BufferTools.emptyFutureBuffer
+    override def isEmpty: Boolean = true
+  }
+
+  /** Construct a [[BodyReader]] with exactly one chunk of data
+    *
+    * This method takes ownership if the passed `ByteBuffer`: any changes to the underlying
+    * buffer will be visible to the consumer of this `MessageBody` and vise versa.
```
rossabaker: *vice versa

http/src/main/scala/org/http4s/blaze/http/BodyReader.scala
ChristopherDavenport: I like `isExhausted`, this will solve a lot of problems if it can be exposed correctly for Http4s.

http/src/main/java/org/http4s/blaze/http/parser/BodyAndHeaderParser.java
```diff
@@ -386,6 +387,7 @@ else if (ch == HttpTokens.LF) {
                     final int buff_size = in.remaining();
 
                     if (remaining_chunk_size <= buff_size) {
+                        // End of this chunk
```
ChristopherDavenport: Pardon my foolishness here, and tangential nature. But doesn't it seem like this method is in dire need of more documentation.

This is literally how we handle all chunked content which is now rather common, and it seems our documentation is a few in line developer notes. `chunkedContent`.

Also how much is the performance benefit of keeping this in Java over Translating to equivalent Scala?

bryce-anderson: It could certainly use more docs: do you want that as part of this PR? As far as converting Java to Scala: the performance difference is not huge, but meaningful, and in many ways the state machine is much easier to deal with in Java since we have break etc. Is there a benefit from moving it from Java to Scala? We can entertain rewriting it in a different change if you like.

http/src/main/scala/org/http4s/blaze/http/BodyReader.scala
bryce-anderson: +1, `isExhausted` sounds great.

http/src/main/java/org/http4s/blaze/http/parser/BodyAndHeaderParser.java
rossabaker: 😆 I was so disoriented by the non-FP code, it didn't even register with me that it was Java when I read it.

ChristopherDavenport: I don't think it needs to be a part of this PR, hence why I stated it was tangential. It just came to my attention.

The docs were just something that when trying to analyze a state machine clear documentation is a must. Otherwise it's almost an attempt to climb into another developers mind.

The benefits I seeing moving to Scala is considering the interesting developments with ScalaJS and ScalaNative. Which are far away from being able to be integrated with blaze, but may be interesting if ScalaNative makes progress. So if we were capable of shifting to pure Scala, that could be a beneficial move.

bryce-anderson:
I agree ScalaNative could be interesting. I'm not opposed to it in the
future. I'll see what I can do as far as docs.

PR #88, opened 2017-08-05
Refactor the HttpServer types

bryce-anderson:
Move some server related stuff to a new package and use the new BodyReader instead of MessageBody.


bryce-anderson:
One thing to keep in mind is that Http2ServerStage.scala will be replaced wholesale in a future commit: it was fixed up here just so that it compiles and tests since it wasn't that difficult to do.

bryce-anderson:
I'd like to merge this evening. I don't think that should be a problem: IMO this is a pretty uncontroversial PR.

PR #89, opened 2017-08-15
New HTTP client implementation

bryce-anderson:
The new client is better than the old in a number of ways.
- The implementation is designed to work in a session pool
- It can do proper streaming on the request body
- Its broken down into smaller, more testable chunks
- Supports HEAD requests and responses


bryce-anderson:
The important parts to watch on this are the ClientStage which has changed almost completely and the body encoders which are new but not all that difficult.

rossabaker:
Note: not ignoring this.  Rushing to get the RCs out the door first.

ChristopherDavenport:
My apologies @bryce-anderson What @rossabaker just said.

With larger changes, don't want to do it an injustice by a weak review

bryce-anderson:
Ha, don't worry: it takes me roughly a day to scroll through this one 😉, I didn't expect to get much meaningful feedback for a bit. I was just trying to be helpful as to what I think is important to pay attention to. There are a number of files that were affected by removing the refinement from `BadMessage` -> `BadRequest` | `BadResponse` which I hope isn't controversial: it made things quite a bit simpler and I don't think there is any real lose of utility.

bryce-anderson:
I'm going to merge this tonight barring objections. Its a large change, so I don't expect it to be perfect, but I'm confident its better than the current state of master and we can continue to fix things if problems are found post-facto.

rossabaker:
I've been traveling so I didn't get a chance to look yet, but I'm confident in your confidence and will still look post facto.

bryce-anderson:
Sounds good, thanks @rossabaker!

PR #90, opened 2017-08-18
Bump to http4s-websocket-0.2.0

rossabaker:
MiMa says it's binary compatible. This will eliminate an irritating eviction warning on http4s >= 0.16 if we release this as 0.12.7, and bump blaze in http4s.

/cc @ZizhengTai @cquiroz

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -110,7 +110,7 @@ lazy val dependencies = Seq(
 )
 
 lazy val specs2              = "org.specs2"                 %% "specs2-core"         % "3.8.6"
-lazy val http4sWebsocket     = "org.http4s"                 %% "http4s-websocket"    % "0.1.6"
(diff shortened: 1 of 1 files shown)


bryce-anderson:
I think that technically it's an unused dependency for blaze.

ZizhengTai:
@bryce-anderson So... we can remove it?

rossabaker:
It's still used in 0.12, isn't it?

bryce-anderson:
Sorry @rossabaker, I didn't notice that it was targeting the 0.12.x branch. Yes, it is still in use there.

rossabaker:
I agree we can drop this in 0.13. I'll do that in a merge up.

I'd like to release this as 0.12.7, along with any trace logging we think can help in the SSL stage.

PR #91, opened 2017-08-23
Trace time spent in various SSLStage methods

rossabaker:
Most of this diff is regrettable reindentation.


bryce-anderson:
Looks good to me. I know this is an effort to help @etorreborre debug his client latency issues, which is good enough, but have we had any other reports of high latency under similarly suspicious conditions?

rossabaker:
He's the first to report it.  I'd still be interested to see a profiler, and whether we're running into contention on the handshake queue.

This logging shouldn't cost anything but System.nanoTime and logger.isTraceEnabled.

rossabaker:
Is there any reason we trace by default?  That seems to make Travis cranky.

bryce-anderson:
No, please feel free to change the log level. Haven't checked, but I vaguely remember this being changed on master.

PR #92, opened 2017-08-30
Add pooled ClientSessionManagerImpl

bryce-anderson:
A rough example of a pooling session manager. I don't think this should be considered 'production ready' since it doesn't put limits on the number of cached sessions and doesn't do a great job cleaning up bad sessions, but it should be a reasonable starting point and provides a mechanism for demonstrating the HTTP/2 client.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
@@ -1,0 +1,167 @@
+package org.http4s.blaze.http
+
+import org.http4s.blaze.channel.nio2.ClientChannelFactory
+import org.http4s.blaze.http.ClientSessionManagerImpl._
+import org.http4s.blaze.http.HttpClientSession.{Closed, ReleaseableResponse, Status}
+import org.http4s.blaze.http.http1.client.Http1ClientStage
+import org.http4s.blaze.http.util.UrlTools.UrlComposition
+import org.http4s.blaze.pipeline.stages.SSLStage
+import org.http4s.blaze.pipeline.{Command, LeafBuilder}
+import org.http4s.blaze.util.Execution
+import org.log4s.getLogger
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Future, Promise}
+import scala.util.control.NonFatal
+import scala.util.{Failure, Success}
+
+private object ClientSessionManagerImpl {
+  case class ConnectionId(scheme: String, authority: String)
+
+  case class Http1SessionProxy(id: ConnectionId, parent: Http1ClientSession) extends Http1ClientSession {
+    override def dispatch(request: HttpRequest): Future[ReleaseableResponse] = parent.dispatch(request)
+    override def status: Status = parent.status
+    override def close(within: Duration): Future[Unit] = parent.close(within)
+  }
+
+  def apply(config: HttpClientConfig): ClientSessionManagerImpl = {
+    val sessionCache = new java.util.HashMap[ConnectionId, java.util.Collection[HttpClientSession]]
+    new ClientSessionManagerImpl(sessionCache, config)
+  }
+}
+
+private final class ClientSessionManagerImpl(sessionCache: java.util.Map[ConnectionId, java.util.Collection[HttpClientSession]], config: HttpClientConfig) extends ClientSessionManager {
+
+  private[this] def getId(composition: UrlComposition): ConnectionId =
+    ConnectionId(composition.scheme, composition.authority)
+
+  private[this] implicit def ec = Execution.trampoline
+
+  private[this] val logger = getLogger
+  private[this] val socketFactory = new ClientChannelFactory(group = config.channelGroup)
+
+  override def acquireSession(request: HttpRequest): Future[HttpClientSession] = {
+    logger.debug(s"Acquiring session for request $request")
+     UrlComposition(request.url) match {
+      case Success(urlComposition) =>
+        val id = getId(urlComposition)
+        val session = findExistingSession(id)
+
+        if (session == null) createNewSession(urlComposition, id)
+        else {
+          logger.debug(s"Found hot session for id $id: $session")
+          Future.successful(session)
+        }
+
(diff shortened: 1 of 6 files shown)


http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
```diff
@@ -0,0 +1,167 @@
...
+  private[this] def findExistingSession(id: ConnectionId): HttpClientSession = sessionCache.synchronized {
+    sessionCache.get(id) match {
+      case null => null // nop
+      case sessions =>
+        var session: HttpClientSession = null
+        val it = sessions.iterator
+        while(session == null && it.hasNext) it.next() match {
+          case h2: Http2ClientSession if h2.status == Closed =>
+            h2.closeNow() // make sure its closed
+            it.remove()
```
rossabaker: This isn't going to kill you with ConcurrentModificationException?  I never mastered when you could remove from an iterator in Java.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
```diff
@@ -0,0 +1,167 @@
...
+          logger.debug(s"Found hot session for id $id: $session")
+          Future.successful(session)
+        }
+
+      case Failure(t) => Future.failed(t)
+    }
+  }
+
+  // WARNING: can emit `null`. For internal use only
+  private[this] def findExistingSession(id: ConnectionId): HttpClientSession = sessionCache.synchronized {
```
rossabaker: What are we getting out of a synchronized map instead of a concurrent map?  Is it the side effects that might be performed in the collection under the key?

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
```diff
@@ -0,0 +1,167 @@
...
+  private[this] def createNewSession(urlComposition: UrlComposition, id: ConnectionId): Future[HttpClientSession] = {
+    logger.debug(s"Creating new session for id $id")
+    val p = Promise[HttpClientSession]
+
+    socketFactory.connect(urlComposition.getAddress).onComplete {
+      case Failure(e) => p.tryFailure(e)
+      case Success(head) =>
+        val clientStage = new Http1ClientStage(config)
+        var builder = LeafBuilder(clientStage)
+        if (urlComposition.scheme == "https") {
```
rossabaker: equalsIgnoreCase, or is this forced lowercase elsewhere?

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
```diff
@@ -0,0 +1,167 @@
...
+    * either cache the session for future use or close it.
+    */
+  override def returnSession(session: HttpClientSession): Unit = {
+    logger.debug(s"Returning session $session")
+    session match {
+      case _ if session.isClosed => () // nop
+      case _: Http2ClientSession => () // nop
+      case h1: Http1ClientSession if !h1.isReady =>
+        logger.debug(s"Closing unready session $h1")
+        h1.closeNow() // we just orphan the Future. Don't care.
```
rossabaker: Are you throwing away an exception if it fails?

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
```diff
@@ -0,0 +1,167 @@
...
+        ()
+
+      case proxy: Http1SessionProxy => addSessionToCache(proxy.id, proxy)
+      case other => sys.error("The impossible happened!")
```
rossabaker: Show us the `other`?

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
bryce-anderson: Nope, not that I recall, and it's better to do so to prevent regressions anyway. I'll change it.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
bryce-anderson: Yes, I didn't think it would be valuable. I can log on the failure pathway if you like: could be interesting to know.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
bryce-anderson: Agreed.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
bryce-anderson: The whole map and it's contents are only accessed from within the lock so it should be safe.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
bryce-anderson: It seemed easier to reason about than having a concurrent map full of concurrent lists.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
rossabaker: Okay, I think this is fine because you're removing from the iterator itself.  If I remember my Java collections right, you die when you remove directly from the collection while iterating, even in a single thread.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
rossabaker: Turning `Future[Unit]` into `Unit` is essentially an empty catch block.  I have been burned rarely, but badly, by this.  If you're going to discard it, at least log the failures.

(I wouldn't have noticed it if not for your comment.)

bryce-anderson:
@rossabaker, do you have any more feedback or should I launch this one?

PR #93, opened 2017-09-01
cleanup examples

bryce-anderson:


examples/src/main/scala/org/http4s/blaze/examples/Http1ServerExample.scala
```diff
@@ -38,7 +39,7 @@ class Http1ServerExample(factory: ServerChannelGroup, port: Int)
 /** Opens a demo server on ports 8080 */
 object NIO1HttpServer {
   def main(args: Array[String]): Unit = {
-    val f = NIO1SocketServerGroup.fixedGroup(workerThreads = Consts.poolSize)
+    val f = NIO1SocketServerGroup.fixedGroup(workerThreads = channel.defaultPoolSize)
```
rossabaker: Constants are usually capitalized.  TitleCase vs UPPER_SNAKE_CASE is inconsistent, but this looks off as camelCase.

bryce-anderson: Good change, so I've thrown it in.

PR #94, opened 2017-09-02
Add some basic types necessary for HTTP/2

bryce-anderson:
Just primitives and helpers.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/Http2Exception.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/Http2Exception.scala
@@ -1,0 +1,109 @@
+package org.http4s.blaze.http.http2
+
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets.UTF_8
+import scala.collection.mutable
+import scala.util.control.NoStackTrace
+
+sealed abstract class Http2Exception(msg: String) extends Exception(msg) with NoStackTrace {
+
+  // A 32-bit unsigned Integer
+  def code: Long
+
+  final def name: String = Http2Exception.errorName(code)
+
+  /** Convert this exception to a stream exception
+    *
+    * @note If this is already a stream exception but with a different stream id, the id will be changed
+    */
+  final def toStreamException(streamId: Int): Http2StreamException = this match {
+    case ex: Http2StreamException if ex.stream == streamId => ex
+    case ex => new Http2StreamException(streamId, ex.code, ex.getMessage)
+  }
+
+  /** Convert this exception to a session exception */
+  final def toSessionException(): Http2SessionException = this match {
+    case Http2StreamException(_, code, msg) => Http2SessionException(code, msg)
+    case ex: Http2SessionException => ex
+  }
+
+  /** Was the exception due to refusal by the peer.
+    *
+    * These exceptions are safe to automatically retry even if the HTTP method
+    * is not an idempotent method. See https://tools.ietf.org/html/rfc7540#section-8.1.4
+    * for more details.
+    */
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/http4s/blaze/http/http2/Http2Exception.scala
```diff
@@ -0,0 +1,112 @@
...
+        val stream = ex match {
+          case ex: Http2StreamException => Some(ex.stream)
+          case _ => None
+        }
+        Some(stream -> ex.getMessage)
+      }
+      else None
+    }
+
+    def unapply(code: Int): Option[Unit] = {
```
rossabaker: Minor, maybe not wanted, but don't forget `unapply` can return `Boolean`.

http/src/main/scala/org/http4s/blaze/http/http2/Http2Result.scala
```diff
@@ -0,0 +1,23 @@
+package org.http4s.blaze.http.http2
+
+/** Result type of many of the codec methods */
+sealed trait Http2Result
```
rossabaker: These are probably not ever in a collection, but extending `Product with Serializable` in the trait improves inference.  I've tried to get into that habit everywhere.

http/src/main/scala/org/http4s/blaze/http/http2/Http2Settings.scala
```diff
@@ -0,0 +1,224 @@
...
+  val HEADER_TABLE_SIZE      = makeKey(0x1, "SETTINGS_HEADER_TABLE_SIZE")
+  val ENABLE_PUSH            = makeKey(0x2, "SETTINGS_ENABLE_PUSH")
+  val MAX_CONCURRENT_STREAMS = makeKey(0x3, "SETTINGS_MAX_CONCURRENT_STREAMS")
+  val INITIAL_WINDOW_SIZE    = makeKey(0x4, "SETTINGS_INITIAL_WINDOW_SIZE")
+  val MAX_FRAME_SIZE         = makeKey(0x5, "SETTINGS_MAX_FRAME_SIZE")
+  val MAX_HEADER_LIST_SIZE   = makeKey(0x6, "SETTINGS_MAX_HEADER_LIST_SIZE")
+}
+
+/** Internal mutable representation of the [[Http2Settings]] */
+private final case class MutableHttp2Settings private(
```
rossabaker: What are we getting out of being a case class if it's mutable?

http/src/main/scala/org/http4s/blaze/http/http2/bits.scala
```diff
@@ -0,0 +1,64 @@
+package org.http4s.blaze.http.http2
+
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+
+object bits {
```
rossabaker: Private so you can change it later?

http/src/main/scala/org/http4s/blaze/http/http2/Http2Exception.scala
bryce-anderson: TIL.

http/src/main/scala/org/http4s/blaze/http/http2/Http2Settings.scala
bryce-anderson: Probably initially a good `toString`, but nothing that I can think of now. I'll remove it.

http/src/main/scala/org/http4s/blaze/http/http2/bits.scala
bryce-anderson: Good call.

PR #95, opened 2017-09-02
Fix flaky test

bryce-anderson:
CI jobs often have a problem with a timer thread
not getting scheduled within a timeout. This makes
the offending test depend on a CountdownLatch so
we don't have to sleep the thread for longer than
necessary.


bryce-anderson:
Since this is a pretty trivial change I'm going to cowboy it in.

PR #98, opened 2017-09-16
Add Header encoder and decoders

bryce-anderson:
In HTTP/2 headers are transmitted using the HPACK compression scheme. We use the twitter HPACK implementation under the hood and wrap them in our own classes to provide a convenient interface and allow us to move to a new HPACK implementation in the future.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
@@ -1,0 +1,91 @@
+package org.http4s.blaze.http.http2
+
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets.US_ASCII
+
+import org.http4s.blaze.http.http2.Http2Exception._
+import org.http4s.blaze.http.http2.Http2Settings.DefaultSettings
+import org.http4s.blaze.util.BufferTools
+import com.twitter.hpack.{Decoder, HeaderListener}
+
+import scala.collection.immutable.VectorBuilder
+
+/** Decoder of HEADERS frame payloads into a collection of key-value
+  * pairs using a HPACK decoder.
+  *
+  * @param maxHeaderListSize maximum allowed size of the header block.
+  *
+  * "The value is based on the uncompressed size of header fields,
+  *  including the length of the name and value in octets plus an
+  *  overhead of 32 octets for each header field."
+  * https://tools.ietf.org/html/rfc7540#section-6.5.2
+  * @param maxTableSize maximum compression table to maintain
+  */
+final class HeaderDecoder(
+  maxHeaderListSize: Int,
+  discardOverflowHeaders: Boolean,
+  val maxTableSize: Int
+) {
+  require(maxTableSize >= DefaultSettings.HEADER_TABLE_SIZE)
+
+  private[this] val acc = new VectorBuilder[(String, String)]
+  private[this] var leftovers: ByteBuffer = null
+  private[this] var headerBlockSize = 0
+
+  private[this] val decoder = new Decoder(maxHeaderListSize, maxTableSize)
+  private[this] val listener = new HeaderListener {
+    override def addHeader(name: Array[Byte], value: Array[Byte], sensitive: Boolean): Unit = {
+      headerBlockSize += 32 + name.length + value.length
+      if (!discardOverflowHeaders || !headerListSizeOverflow) {
+        acc += new String(name, US_ASCII) -> new String(value, US_ASCII)
+        ()
+      }
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
```diff
@@ -0,0 +1,91 @@
+package org.http4s.blaze.http.http2
+
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets.US_ASCII
+
+import org.http4s.blaze.http.http2.Http2Exception._
+import org.http4s.blaze.http.http2.Http2Settings.DefaultSettings
+import org.http4s.blaze.util.BufferTools
+import com.twitter.hpack.{Decoder, HeaderListener}
```
rossabaker: Is this a new dependency?  I didn't notice it sneaking in.  I'm not too fussed about it since it's slated for removal.

http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
```diff
@@ -0,0 +1,91 @@
...
+        ()
+      }
+    }
+  }
+
+  /** Set the HEADER_TABLE_SIZE parameter */
+  def setMaxHeaderTableSize(max: Int): Unit = decoder.setMaxHeaderTableSize(max)
+
+  /** Returns the header collection and clears the builder */
+  def result(): Seq[(String,String)] = {
```
rossabaker: leftovers is kept for a subsequent run, right?

http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
```diff
@@ -0,0 +1,91 @@
...
+        b.put(buff).flip()
+        leftovers = b
+      }
+
+      if (endHeaders) {
+        decoder.endHeaderBlock()
+      }
+
+      Continue
+    } catch { case t: Throwable =>
```
rossabaker: I'd do NonFatal here so you're not catching JVM errors.

http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
```diff
@@ -0,0 +1,91 @@
...
+) {
+  require(maxTableSize >= DefaultSettings.HEADER_TABLE_SIZE)
+
+  private[this] val acc = new VectorBuilder[(String, String)]
+  private[this] var leftovers: ByteBuffer = null
+  private[this] var headerBlockSize = 0
+
+  private[this] val decoder = new Decoder(maxHeaderListSize, maxTableSize)
+  private[this] val listener = new HeaderListener {
+    override def addHeader(name: Array[Byte], value: Array[Byte], sensitive: Boolean): Unit = {
```
rossabaker: What does `sensitive` mean?  Who is passing it to you?  I'm surprised this isn't a function of the name.

http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
bryce-anderson: No, it had always been there. Unfortunately it's no longer maintained as it was folded into the Netty project. That said, it still works as expected so there is little stunning force to migrate to a new one.

http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
bryce-anderson: Hpack allows you to flag a header as sensitive to signal that it shouldn't be cashed. Unfortunately there isn't a great api to forward that info.

http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
bryce-anderson: It should always be empty after we've gotten to the end of the header block and want to call result. Maybe this could be made better by changing the name to `finish` and enforcing that it has been called after `decode(.., endHeaders = true)`.

bryce-anderson:
Thanks for the feedback, @rossabaker.

http/src/main/scala/org/http4s/blaze/http/http2/HeaderDecoder.scala
rossabaker: Okay. My concern was that it wasn't being cleared when the rest of the state was.

bryce-anderson: Yeah, and it's a good call just to clear it: who knows if the hpack decoder could have left some stay padding laying around, so good eye.

PR #103, opened 2017-09-26
Add HeaderAggregatingFrameListener and Http2FrameSerializer

bryce-anderson:


bryce-anderson:
I debated splitting this but the testing is co-dependent. If you prefer, I can break it up and comment out the tests for the first chunk.

http/src/main/scala/org/http4s/blaze/http/http2/Http2FrameSerializer.scala
```diff
@@ -0,0 +1,293 @@
...
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+
+import org.http4s.blaze.http.http2.Http2Settings.Setting
+import org.http4s.blaze.util.BufferTools
+
+private object Http2FrameSerializer {
+  import bits._
+
+  // Override the scala provided `require(condition, => msg)` to avoid the thunks
```
rossabaker: Is this more important than getting the values (e.g., line 29)?  Could we have our cake and eat it too with a *gasp* macro?

http/src/main/scala/org/http4s/blaze/http/http2/Http2FrameSerializer.scala
```diff
@@ -0,0 +1,293 @@
...
+    writeFrameHeader(PingSize, FrameTypes.PING, flags, 0, buffer)
+    buffer
+      .put(data)
+      .flip()
+
+    buffer
+  }
+
+  def mkGoAwayFrame(lastStreamId: Int, error: Http2Exception): Seq[ByteBuffer] = {
+    val msgString = error.getMessage.getBytes(StandardCharsets.UTF_8)
```
rossabaker: Are these always UTF-8?  Most of the time the standards refer to US-ASCII, although it's probably going to be the same, and this is a newer spec.  This is probably right, but just asking.

http/src/test/scala/org/http4s/blaze/http/http2/CodecUtils.scala
```diff
@@ -6,8 +6,11 @@ import java.util
 import org.http4s.blaze.util.BufferTools._
 
 object CodecUtils {
+
+  def byteData: Array[Byte] = "The quick brown fox jumps over the lazy dog".getBytes()
```
rossabaker: Break the habit of using the platform-default encoding for `getBytes`.  It's just a test, but young programmers may be watching and pasting at any time.

http/src/test/scala/org/http4s/blaze/http/http2/HeaderAggregatingFrameListenerSpec.scala
```diff
@@ -0,0 +1,265 @@
+package org.http4s.blaze.http.http2
+
+import java.nio.ByteBuffer
+
+import org.http4s.blaze.http._
+import org.http4s.blaze.util.BufferTools
+import org.specs2.mutable.Specification
+
+class HeaderAggregatingFrameListenerSpec extends Specification {
```
rossabaker: Do you think this could benefit from property testing?  Especially when we see "round trip".

http/src/main/scala/org/http4s/blaze/http/http2/Http2FrameSerializer.scala
bryce-anderson: The "additional debug data" of the GOAWAY frame doesn't have a prescribed format.

http/src/test/scala/org/http4s/blaze/http/http2/HeaderAggregatingFrameListenerSpec.scala
bryce-anderson: Yes, it likely would.

bryce-anderson:
I'll add the scalacheck tests. I'd thought about it before you mentioned it, but I took the lazy route and didn't write them, so thanks of calling me on it.

PR #104, opened 2017-09-27
Don't write an empty Seq of buffers to the socket

bryce-anderson:
The nio2.ByteBufferHead will attempt to write an empty
Seq of buffers which can cause problems on some platforms.
Instead we should short circuit and return successful
immediately.


bryce-anderson:
This will probably need to be cherry picked into other branches as well.

ChristopherDavenport:
Do we have any tests we can implement for this functionality? Working on a Windows build for testing purposes so having a before and after could be useful.

bryce-anderson:
Alright, retargeted to 0.12.x and added some tests. The tests don't care about the change on my macbook, but they could make the difference on Windows.

PR #105, opened 2017-09-27
Add Appveyor Config

ChristopherDavenport:
Basic AppVeyor Config

References http4s/http4s#1426


.appveyor.yml
```diff
@@ -0,0 +1,31 @@
...
+    TEST_COMPILE: test:compile
+    TEST_SUITE: test
+  matrix:
+    - SCALA_VERSION: 2.11.11
+    - SCALA_VERSION: 2.12.3
+
+install:
+  - ps: |
+      Add-Type -AssemblyName System.IO.Compression.FileSystem
+      if (!(Test-Path -Path "C:\sbt\lib\local-preloaded\org.scala-sbt\sbt\0.13.15\jars\sbt.jar" )) {
```
rossabaker: This is probably true on the other PR.  Can we get that repeated 0.13.15 into the global environment and reference it here?  And is there any reason not to go straight to 0.13.16?  I realize the main build isn't sbt-1.0 yet.

ChristopherDavenport: Yes. That's a transition I would like to make. I thought it seemed like a good idea to get a working version implemented and then transition to perfecting it. As I expected a bit more difficulties on rollout.

rossabaker: Alright, I'm happy with this.  Appveyor found one bug in http4s, so I'm curious to see what it finds here.

rossabaker:
Is there any reason not to merge this?

ChristopherDavenport:
Nope once we merge, then I would enable the build.

PR #107, opened 2017-09-30
Specify Charset in `String.getBytes` calls

bryce-anderson:
Not specifying it means you get a platform specific `Charset`.

### DIFF
--- a/http/src/test/scala/org/http4s/blaze/http/parser/ServerParserSpec.scala
+++ b/http/src/test/scala/org/http4s/blaze/http/parser/ServerParserSpec.scala
@@ -215,7 +215,7 @@ class ServerParserSpec extends Specification {
 
     "Parse a full request" in {
       val p = new Parser()
-      val b = ByteBuffer.wrap(mockFiniteLength.getBytes())
+      val b = strToBuffer(mockFiniteLength)
 
       p.parseLine(b) should_== (true)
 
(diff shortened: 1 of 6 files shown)


http/src/test/scala/org/http4s/blaze/http/parser/ServerParserSpec.scala
```diff
@@ -215,7 +215,7 @@ class ServerParserSpec extends Specification {
 
     "Parse a full request" in {
       val p = new Parser()
-      val b = ByteBuffer.wrap(mockFiniteLength.getBytes())
+      val b = strToBuffer(mockFiniteLength)
```
ChristopherDavenport: Just confirming the use of [`ISO_8859_1`](https://github.com/http4s/blaze/blob/de3395f0dc02076bd1cd806346ae4eabd9c8ad52/http/src/test/scala/org/http4s/blaze/http/parser/ServerParserSpec.scala#L14) as opposed to `UTF-8` implemented above.

bryce-anderson: It should be fine as all the characters in the tests are ASCII and both UTF8 and iso-8859-1 are supersets of ASCII. That said, I'm not thrilled with the smattering of different ASCII superset charsets I've used in different tests over the years.

Issue #97, opened 2017-09-13
intermittent failures in 2.12 community build: 


SethTisue:
seen at https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/1906/consoleFull

```
[blaze] [error] Failed tests:
[blaze] [error] 	org.http4s.blaze.util.ActorSpec
[blaze] [error] Error during tests:
[blaze] [error] 	org.http4s.blaze.pipeline.stages.SerializingStageSpec
```

SethTisue:
`ActorSpec` failure seen again at https://scala-ci.typesafe.com/view/All/job/scala-2.12.x-integrate-community-build/1953/consoleFull

SethTisue:
`SerializingStageSpec` failed again: https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-integrate-community-build/1997/consoleFull

SethTisue:
`ChannelSpec` failed again at https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/2014/consoleFull

this is happening often enough to be an annoyance for us. I can "solve" it at our end by disabling blaze's test suite entirely, but that's not the ideal solution. do these tests perhaps just need to have longer timeouts or something easy like that...?

rossabaker:
@bryce-anderson, I've been leaving this for you, but I'll take a poke at it tomorrow if you're unavailable.

rossabaker:
What is the execution model on this community build?  A lot of these tests are using a global execution context.  Is it possible any of the other projects that are running could be blocking threads, or do we have better isolation than that?  I don't see an obvious timeout to tweak, but maybe we could provide our own execution context for tests, since in isolation, this all seems to work reliably.

SethTisue:
we only run one project at a time in the community build, so there can't be anything else happening in the same JVM.

but, it's possible for the same machine to be running a couple other jobs at the same time, so usually the only special thing is that the machine may be heavily loaded

often that merely means longer timeouts are needed, but it does sometimes reveal actual bugs that don't tend to crop up except under load

bryce-anderson:
I can poke at this later this evening. Most of those timeouts are pretty long, `SerializationStageSpec` failed with a timeout of 10 seconds. The ActorSpec and ChannelSpec don't involve a timeout, so there seems to be some real work to do on those two.

SethTisue:
(oops)

SethTisue:
```
[...]

PR #110, opened 2017-10-06
ActorSpec should throw the exception that causes it to halt

### LINKED ISSUE #109: In ActorSpec we should be throwing an exception if we fail the flag condition
https://github.com/http4s/blaze/blob/daba0ccb57494641df960fa63556920f784956a5/core/src/test/scala/org/http4s/blaze/util/ActorSpec.scala#L55

bryce-anderson:
Closes #109


bryce-anderson:
Inspired by #97.

PR #112, opened 2017-10-06
catch exceptions in NIO2SocketServerGroup

### LINKED ISSUE #111: Not catching potential exceptions in NIO2SocketServerGroup
https://github.com/http4s/blaze/blob/6e4c16ba3391da1cc7ee15998d0e18ef2dbb2b04/core/src/main/scala/org/http4s/blaze/channel/nio2/NIO2SocketServerGroup.scala#L110

That can throw, and since we're running it in a loop, those exceptions may never see the light of day. This could result in problems if we're shutting down the group and expect shutdown hooks to fire.

bryce-anderson:
Closes #111 

If we don't catch exceptions in AsynchronousSocketChannel
we could lose them since the loop is happening asynchronously.

Reported in #97


bryce-anderson:
@rossabaker, I believe so. This is one of those classic race conditions that can be pretty difficult to test.

PR #114, opened 2017-10-06
Make SerializingStageSpec deterministic

bryce-anderson:
The tests previously made hundreds of threads which
is really poor form.


bryce-anderson:
Work spawned by #97.

bryce-anderson:
They are now arbitrarily slow since they are being driven by the user. This is not ideal as it doesn't test race conditions, but the previous spec hard did that either. Long term, I want to remove this as it's really only a crutch.

bryce-anderson:
I opened #118 to simply remove these in the future.

PR #116, opened 2017-10-06
Fix Typelevel Scala in Compile Stage

ChristopherDavenport:
This should fix the current primary issue with publish stage for master on 2.12.3 and 2.11.11


bryce-anderson:
Thanks for taking care of this @ChristopherDavenport!

PR #120, opened 2017-10-07
Add Suppression Option to Scaladoc Linking

ChristopherDavenport:
I see some issues with suppressing. However I worry how long we have been out of commission.

Other options are looking through what is necessary for upgrading rig.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,9 @@ organization in ThisBuild := "org.http4s"
 lazy val commonSettings = Seq(
   version := "0.14.0-SNAPSHOT",
   description := "NIO Framework for Scala",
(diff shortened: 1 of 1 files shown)


bryce-anderson:
A few of the broken links don't make sense to me: 
```
error] blaze/core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala:48: Could not find any member to link for "Disconnected".
[error]   /** Shuts down the stage, deallocating resources, etc.
[error]   ^
```
The line is [here](https://github.com/http4s/blaze/blob/master/core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala#L50), and the link looks fine to me.

rossabaker:
scaladoc requires a lot of absolute references where values are clearly in scope.

PR #121, opened 2017-10-07
Remove TimingTools

### LINKED ISSUE #119: Spinning time exceed the `Duration` parameter in TimingTools should throw a TimeoutException
https://github.com/http4s/blaze/blob/86773041503c6427b251019e3dcb51dbb45a9353/core/src/test/scala/org/http4s/blaze/util/TimingTools.scala#L12

This may be the cause of the failures being reported in #97, eg the test is just taking more than 5 seconds and spinning decides to complete silently.

bryce-anderson:
Fixes #119

TimingTools.spin is not a great idea and pretty much
everything you could want to spin on can be done with
a CountDownLatch.


bryce-anderson:
Merging this one since it only affects tests and, at least IMO, is obviously better.

rossabaker:
I'm travelling. I'll try to explain post facto review tomorrow.

On Oct 9, 2017 18:58, "Bryce Anderson" <notifications@github.com> wrote:

Merging this one since it only affects tests and, at least IMO, is
obviously better.

—
You are receiving this because your review was requested.

Reply to this email directly, view it on GitHub
<https://github.com/http4s/blaze/pull/121#issuecomment-335320788>, or mute
the thread

PR #123, opened 2017-11-02
Dependency bumps

rossabaker:


rossabaker:
I left specs2 behind.  Just looking for quick wins.  I want to release a blaze-0.12.10 to get rid of eviction warnings in http4s.

PR #124, opened 2017-11-10
Add build info

rossabaker:
This will let us print the blaze version like we do the Tomcat and Jetty versions in http4s/http4s#1536.


ChristopherDavenport:
Sometimes it appears appveyor doesn't even try.

rossabaker:
We've been getting yellow responses on Travis builds that finished long ago, too.

PR #125, opened 2017-11-19
Introduce misc tools and the SessionCore

bryce-anderson:
The SessionCore is the most notable part, and is
essentially a glue type that contains references
to the different components of the session. This
allows each component to reach other components
since the SessionCore is often passed as a constructor
parameter. It may not live forever, but is a handy
abstraction for now.


bryce-anderson:
I don't think that appveyor is setup for this branch, but Travis is 👍

rossabaker:
Not ignoring this, just a very busy weekend.  I should be able to look this afternoon.

http/src/main/scala/org/http4s/blaze/http/http2/FlowStrategy.scala
```diff
@@ -20,16 +20,20 @@ trait FlowStrategy {
...
     */
-  def checkStream(session: SessionFlowControl, stream: StreamFlowWindow): Increment
+  def checkStream(stream: StreamFlowWindow): Increment
 }
 
 object FlowStrategy {
   final class Increment private[FlowStrategy](val session: Int, val stream: Int) {
-    override def toString: String = s"Increment($session, $stream)"
+    override def toString = s"Increment($session, $stream)"
+    override def hashCode(): Int = session + stream
```
rossabaker: This seems like a weird hash code.  Is it because these values are fairly random anyway?

Any reason this isn't just a case class?

http/src/main/scala/org/http4s/blaze/http/http2/Http2StageTools.scala
```diff
@@ -0,0 +1,56 @@
...
+  // Response pseudo header
+  val Status = ":status"
+  val TE = "te"
+  val Connection = HeaderNames.Connection
+  val ContentLength = HeaderNames.ContentLength
+
+  /** Check of the header name is a valid http2 header name.
+    * Pseudo headers are considered invalid and should be screened before hand.
+    */
+  def validHeaderName(str: String): Boolean = {
```
rossabaker: Is this faster than `str.nonEmpty && str.forall(...)`?  (Probably, but you reinvented a wheel.)

http/src/main/scala/org/http4s/blaze/http/http2/FlowStrategy.scala
bryce-anderson: It's only so I have something to go with the overridden equals. It's not a case class because I wanted to implement the `apply` method, or more specifically, I didn't want the synthesized `apply` so I could optimize the common case of `Empty`. Maybe that's not a great reason: I also dislike the noise.

http/src/main/scala/org/http4s/blaze/http/http2/Http2StageTools.scala
bryce-anderson: Won't the `String.forall` result in boxing each `Char`? That is what I was trying to avoid.

PR #126, opened 2017-11-30
Add StreamIdManager

bryce-anderson:
Introduce the accounting abstraction for stream ids.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/StreamIdManager.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/StreamIdManager.scala
@@ -1,0 +1,92 @@
+package org.http4s.blaze.http.http2
+
+private object StreamIdManager {
+
+  /** Create a new [[StreamIdManager]] */
+  def apply(isClient: Boolean): StreamIdManager= create(isClient, 0)
+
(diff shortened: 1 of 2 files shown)


bryce-anderson:
I'll plan on merging this sometime this weekend unless there are objections.

http/src/main/scala/org/http4s/blaze/http/http2/StreamIdManager.scala
```diff
@@ -0,0 +1,93 @@
+package org.http4s.blaze.http.http2
+
+private object StreamIdManager {
+
+  /** Create a new [[StreamIdManager]] */
+  def apply(isClient: Boolean): StreamIdManager
+    = create(isClient, 0)
```
rossabaker: This is an odd indentation of the `=`.

bryce-anderson:
@rossabaker, no worries, I appreciate the review.

http/src/main/scala/org/http4s/blaze/http/http2/StreamIdManager.scala
bryce-anderson: Indeed, I'll fix it.

PR #128, opened 2017-12-02
Introduce the PingManager and WriteInterest and WriteController interfaces

bryce-anderson:
Also moved the mocks to their own package.


bryce-anderson:
Other than the `PingManager` which I don't expect to be controversial, the rest of this PR is refactor and and private interfaces, so I'm going to try and get this in by tomorrow evening to keep the ball rolling. Still nearly ~4k lines of code to merge 😄.

If this gets merged before you get a chance to review, feedback still welcome!

http/src/main/scala/org/http4s/blaze/http/http2/PingManager.scala
```diff
@@ -1,4 +1,58 @@
...
+
+  private[this] val logger = getLogger
+  private[this] var currentPing: Option[PingState] = None
+
+  // Must be called from within the sessionExecutor
+  def ping(): Future[Duration] = currentPing match {
+    case Some(_) =>
+      val msg = "Ping already in progress"
+      val ex = new IllegalStateException(msg)
+      logger.warn(ex)(msg)
```
rossabaker: I'm not generally a huge fan of logging and passing an exception.  Makes it tend to appear twice when it happens once.

http/src/main/scala/org/http4s/blaze/http/http2/PingManager.scala
```diff
@@ -1,4 +1,58 @@
...
+        currentPing = None
+
+        if (ByteBuffer.wrap(data).getLong != sent) { // data guaranteed to be 8 bytes
+          val msg = "Received ping response with unknown data."
+          val ex = new Exception(msg)
+          logger.warn(ex)(msg)
+          continuation.tryFailure(ex)
+          ()
+        } else {
+          val duration = Duration.fromNanos((System.currentTimeMillis - sent) * 1000000)
```
rossabaker: There's a `Duration.create(Long, TimeUnit)`.  I haven't checked the innards to see if that's faster or not.

http/src/main/scala/org/http4s/blaze/http/http2/SessionCore.scala
```diff
@@ -12,12 +12,11 @@ import scala.concurrent.duration.Duration
   */
 private trait SessionCore {
   // Fields
-
   def serialExecutor: ExecutionContext
 
   def localSettings: Http2Settings // The settings of this side
 
-  def remoteSettings: Http2Settings // The peers settings.
+  def remoteSettings: MutableHttp2Settings // The peers settings.
```
rossabaker: peer's

http/src/main/scala/org/http4s/blaze/http/http2/PingManager.scala
bryce-anderson: I bet its pretty quick: it appears to be an alias for the constructor. 😄 I'll definitely use `create`.

http/src/main/scala/org/http4s/blaze/http/http2/PingManager.scala
bryce-anderson: Thats reasonable. Do you no like logging at debug level as well? I think this is debug worthy at a minimum, and it's not possible to control whether the passed exception actually gets logged.

rossabaker: I think that's a fair compromise if you're worried about the caller not logging it.  I just tend think if you pass it you pass off responsibility, and get annoyed when I see twice as many exceptions as I have problems.

bryce-anderson: This is only used privately so the details are completely in our hands here, so I'm convinced to not log here.

PR #130, opened 2017-12-03
Add WriteControllerImpl

bryce-anderson:


bryce-anderson:
This PR will ultimately be merged into the `merge-new-client-http2` branch, but is dependent on the `merge-new-client-http2-PingManager` branch, so thats where it's targeted for now to make it easy to read.

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
```diff
@@ -0,0 +1,154 @@
...
+import org.log4s.getLogger
+
+import scala.collection.mutable.ArrayBuffer
+import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Success}
+import scala.util.control.NonFatal
+
+/** Gracefully coordinate writes
+  *
+  * The `WriteController`s job is to direct outbound data in both a fair, efficient, and
```
rossabaker: Missing apostrophe after `WriteController`

"both a (3 things)"

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
```diff
@@ -0,0 +1,154 @@
...
+
+          case Flushing =>
+            state = Idle
+
+          case Closing(_) if pendingInterests =>
+            doWrite()
+
+          case Closing(p) =>
+            state = Closed
+            p.trySuccess(())
```
rossabaker: What if it doesn't succeed?

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
```diff
@@ -0,0 +1,154 @@
...
+
+          case Closed =>
+            () // shouldn't get here
```
rossabaker: If you shouldn't and did, you should log it.

http/src/test/scala/org/http4s/blaze/http/http2/WriteControllerImplSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+import org.specs2.mutable.Specification
+
+import scala.collection.mutable
+import scala.concurrent.{Future, Promise}
+
+class WriteControllerImplSpec extends Specification {
+
+  private def mockData(start: Int): ByteBuffer =
+    ByteBuffer.wrap {
+    (0 until (start + 1)).map(_.toByte).toArray
```
rossabaker: `(0 to start)`?

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
bryce-anderson: It should never fail. I"ll make it the standard `.success(..)`.

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
bryce-anderson: I'll throw an exception instead since there is a serious bug if we get here. The sessions `serialExecutor` will tear everything down on unhandled exceptions and make a fuss.

http/src/test/scala/org/http4s/blaze/http/http2/WriteControllerImplSpec.scala
bryce-anderson: `start` is an awful name. It needs to be at least size 1 or else it gets filtered. I've just changed it to `i`.

bryce-anderson:
Going to reopen as a new PR since GH seems to have not been able to handle the branch retarget.

PR #131, opened 2017-12-03
Add WriteControllerImpl

bryce-anderson:


bryce-anderson:
This is a continuation of #130.

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
```diff
@@ -0,0 +1,154 @@
...
+import org.log4s.getLogger
+
+import scala.collection.mutable.ArrayBuffer
+import scala.concurrent.{Future, Promise}
+import scala.util.{Failure, Success}
+import scala.util.control.NonFatal
+
+/** Gracefully coordinate writes
+  *
+  * The `WriteController`s job is to direct outbound data in a fair, efficient, and
```
rossabaker: `WriteController`'s (needs apostrophe)

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
```diff
@@ -0,0 +1,154 @@
...
+  session: SessionCore,
+  highWaterMark: Int,
+  tailStage: TailStage[ByteBuffer]
+) extends WriteController {
+  import WriteControllerImpl._
+
+  private[this] val logger = getLogger
+  // TODO: if we implement priorities, we should prioritize writes as well.
+  private[this] val interestedStreams = new java.util.ArrayDeque[WriteInterest]
+  private[this] val pendingWrites = new util.ArrayDeque[Seq[ByteBuffer]]
```
rossabaker: Odd style: one is fully qualified one is not.

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
```diff
@@ -0,0 +1,154 @@
...
+    case Idle | Flushing | Closing(_) =>
+      pendingWrites.addLast(data)
+      maybeWrite()
+      true
+
+    case Closed =>
+      false
+  }
+
+  def write(data: ByteBuffer): Boolean = write(data::Nil)
```
rossabaker: Since you like to save every nanosecond, is that the fastest `Seq` you can create?  I think I've seen this a lot of places.  (I don't know, but it seems like the sort of thing you might.)

http/src/test/scala/org/http4s/blaze/http/http2/WriteControllerImplSpec.scala
```diff
@@ -0,0 +1,149 @@
...
+import org.specs2.mutable.Specification
+
+import scala.collection.mutable
+import scala.concurrent.{Future, Promise}
+
+class WriteControllerImplSpec extends Specification {
+
+  private def mockData(i: Int): ByteBuffer =
+    ByteBuffer.wrap {
+    (0 until (i + 1)).map(_.toByte).toArray
```
rossabaker: I guess my point here was `to` means you don't need to `+ 1`, unless you think `+ 1` is instructive.

bryce-anderson: Oh, I see. I don't necessarily think the +1 is important so I'll change it.

http/src/main/scala/org/http4s/blaze/http/http2/WriteControllerImpl.scala
bryce-anderson: I think it's about as good as you can get for a single element: should just be a single `::`. I'm interested if you know of a better one.

PR #133, opened 2017-12-16
Add StreamState

bryce-anderson:
StreamState is the junction between the H2 session
and the individual dispatches.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/Http2Connection.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/Http2Connection.scala
@@ -47,14 +47,17 @@ trait Http2ClientConnection extends Http2Connection with Http2ClientSession {
 }
 
 object Http2Connection {
-  sealed trait ConnectionState
+  sealed abstract class ConnectionState {
+    final def closing: Boolean = !running
+    final def running: Boolean = this == Running
+  }
 
   /** The `Running` state represents a session that is active and able to accept
     * new streams.
     */
   case object Running extends ConnectionState
 
-  sealed trait Closing extends ConnectionState
+  sealed abstract class Closing extends ConnectionState
 
   /** The `Draining` state represents a session that is no longer accepting new
     * streams and is in the process of draining existing connection.
--- a/http/src/main/scala/org/http4s/blaze/http/http2/Http2Connection.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/Http2Connection.scala
@@ -47,7 +47,7 @@ trait Http2ClientConnection extends Http2Connection with Http2ClientSession {
 }
 
 object Http2Connection {
-  sealed abstract class ConnectionState {
+  sealed abstract class ConnectionState extends Product with Serializable {
     final def closing: Boolean = !running
     final def running: Boolean = this == Running
   }
--- a/http/src/main/scala/org/http4s/blaze/http/http2/OutboundStreamState.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/OutboundStreamState.scala
@@ -1,6 +1,7 @@
 package org.http4s.blaze.http.http2
 
 import scala.concurrent.Promise
+import scala.concurrent.duration.Duration
 
 /** Representation of outbound streams
   *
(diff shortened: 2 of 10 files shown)


http/src/main/scala/org/http4s/blaze/http/http2/Http2Connection.scala
```diff
@@ -47,14 +47,17 @@ trait Http2ClientConnection extends Http2Connection with Http2ClientSession {
 }
 
 object Http2Connection {
-  sealed trait ConnectionState
+  sealed abstract class ConnectionState {
```
rossabaker: This is probably true a lot of places, but do you want to extend Product with Serializable for better inference?

http/src/main/scala/org/http4s/blaze/http/http2/OutboundStreamState.scala
```diff
@@ -0,0 +1,62 @@
...
+  * expect that we will have HEADERS to send right when it is born, so
+  * we need to make the stream ID lazy since they must be used in
+  * monotonically increasing order.
+  */
+private final class OutboundStreamState(session: SessionCore) extends StreamState(session) {
+  private[this] var lazyStreamId: Int = -1
+  private[this] var lazyFlowWindow: StreamFlowWindow = null
+
+  private[this] def initialized: Boolean = lazyStreamId != -1
+  private[this] def uninitializedException: Nothing = {
```
rossabaker: Unimportant because it's private, but without parens on this, I was expecting to find a cached exception until I scrolled back up.

http/src/main/scala/org/http4s/blaze/http/http2/OutboundStreamState.scala
```diff
@@ -0,0 +1,62 @@
...
+      // Before we initialized the stream, we began to drain or were closed.
+      val ex = Http2Exception.REFUSED_STREAM.rst(0, "Session closed before stream was initialized")
+      p.failure(ex)
+      ()
+    } else {
+      session.streamManager.registerOutboundStream(this) match {
+        case Some(streamId) =>
+          // Note: it is assumed that these will both be initialized or neither are initialized
+          lazyFlowWindow = session.sessionFlowControl.newStreamFlowWindow(streamId)
+          lazyStreamId = streamId
```
rossabaker: We're not initialized, so we set lazyStreamId to streamId, which throws if we're not initialized?

bryce-anderson: I've overloaded `StreamState.streamId` with a local value (from the `Some`). This is a pattern I typically love since it makes it impossible to use the incorrect symbol, but I'm starting to think I'm alone in that regard.

Does that solve your question, or have I simply misunderstood it?

http/src/main/scala/org/http4s/blaze/http/http2/OutboundStreamState.scala
bryce-anderson: I can add them.

http/src/main/scala/org/http4s/blaze/http/http2/Http2Connection.scala
bryce-anderson: Is this just to avoid the inference of the common subtype `ConnectionState with Product with Serializable` since all the sub types are case *things*? 
(this is the first time I've heard of this strategy)

But yeah, I can do that if you like.

rossabaker: Yep.  I've started habitually making all the types this to improve the inference.

bryce-anderson:
@rossabaker, thanks for the quick review! I think I've addressed your feedback, but I'm not sure if I understood that question.

http/src/main/scala/org/http4s/blaze/http/http2/OutboundStreamState.scala
rossabaker: You can't reference the wrong one, but shadowing is easy to overlook, as I did.  But yeah, I understand how it works now.  It's fine.

bryce-anderson: I think it is difficult for reviewers (I get dinged for this at $work somewhat frequently) and since good code should be read far more than it's written, I've just changed it for reading clarity.

PR #134, opened 2017-12-22
Merge new client http2 stream manager

bryce-anderson:
Other than introducing the `StreamManager` most of this PR is noise due to refactoring `StreamState` into an interface and an implementation. One important change that it took on I'll point out in a comment. I can put them back in the same file and it make make the PR clear as it won't look like a wholesale replacement if reviewers prefer.


http/src/main/scala/org/http4s/blaze/http/http2/StreamManager.scala
```diff
@@ -1,11 +1,82 @@
...
+    * @return true if the stream existed and was closed, false otherwise
+    */
+  def rstStream(cause: Http2StreamException): Boolean
+
+  /** Called by a `StreamState` to remove itself from the StreamManager
+    *
+    * @param streamState the stream being closed
+    * @return true if the stream existed and was closed, false otherwise
+    */
+  def streamClosed(streamState: StreamState): Boolean
```
bryce-anderson: Instead of `streamFinished` we now have `streamClosed` which returns a `Boolean` to signal whether the stream was still considered active by the session. This puts the onus on `StreamState` to manage its own exceptions appropriately, and it can escalate session level exceptions if needed.

bryce-anderson:
It's really not as large as it looks, the only really interesting change is the `StreamManager`, it just looks that way because the refactor wasn't captured well. Let me see if I can throw some files back together real quick.

bryce-anderson:
Closing in favor of https://github.com/http4s/blaze/pull/135, which is strictly a copy-paste of this diff. After review, I'll fix the non-technical formatting changes.

PR #135, opened 2017-12-22
Merge new client http2 stream manager pretty

bryce-anderson:
Less angry diff of https://github.com/http4s/blaze/pull/134 made only by copying the new `*StreamStateImpl` contents into the original `*StreamState` files so that git can make a ridiculous diff. After review, I'll move the implementations back into their own files and merge.


bryce-anderson:
@rossabaker, this diff does a much better job of capturing the actually changes. Sorry for the noise.

http/src/main/scala/org/http4s/blaze/http/http2/StreamState.scala
```diff
@@ -269,15 +292,32 @@ private abstract class StreamState(session: SessionCore)
...
         case Some(EOF) => None
         case Some(other) =>
           logger.warn(other)(s"Unknown error in stream($streamId)")
           Some(INTERNAL_ERROR.rst(streamId, "Unhandled error in stream pipeline"))
       }
 
-      // Remove ourselves from the streamManager
-      session.streamManager.streamFinished(this, http2Ex)
+      // Remove ourselves from the streamManager. If the stream didn't exist, it is
+      // because this tream was reset by the remote, and thus we shouldn't send a
```
bryce-anderson: s/tream/stream

http/src/main/scala/org/http4s/blaze/http/http2/StreamState.scala
```diff
@@ -4,13 +4,38 @@ import java.nio.ByteBuffer
...
+  def performStreamWrite(): Seq[ByteBuffer]
+
+  /** Called by the session when a DATA frame has been received from the remote peer */
+  def invokeInboundData(endStream: Boolean, data: ByteBuffer, flowBytes: Int): MaybeError
+
+  /** Called by the session when a HEADERS has been received from the remote peer */
+  def invokeInboundHeaders(priority: Priority, endStream: Boolean, headers: Seq[(String,String)]): MaybeError
+
+  /** Close the stream, optionally with an error */
+  def closeWithError(t: Option[Throwable]): Unit
```
rossabaker: Similar question to the other close Unit.

http/src/main/scala/org/http4s/blaze/http/http2/StreamState.scala
```diff
@@ -269,15 +292,32 @@ private abstract class StreamState(session: SessionCore)
...
+          val frame = session.http2Encoder.rstFrame(streamId, ex.code)
+          session.writeController.write(frame)
+          ()
+
+        case Some(ex: Http2SessionException) =>
+          logger.info(s"Stream($streamId) finished with session exception")
+          // TODO: this doens't feel like the responsibility of the StreamState
+          session.invokeShutdownWithError(http2Ex, "streamFinished")
+
+        case _ => () // nop
```
rossabaker: I'd make this a `case None` to preserve exhaustiveness checking in case `http2Ex` ever gets widened.  I don't expect it to, but empty exception handlers in long methods are playing with fire when you refactor.

http/src/main/scala/org/http4s/blaze/http/http2/StreamManager.scala
```diff
@@ -1,11 +1,82 @@
...
+  def handlePushPromise(streamId: Int, promisedId: Int, headers: Headers): Http2Result
+
+  /** Update the specified flow window with the specified bytes
+    *
+    * @note stream ID 0 indicates the session flow window
+    */
+  def flowWindowUpdate(streamId: Int, sizeIncrement: Int): MaybeError
+
+  /** Close the `StreamManager` and all the associated streams immediately */
+  def close(cause: Option[Throwable]): Unit
```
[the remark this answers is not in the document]
bryce-anderson: I didn't add a `Future[Unit]` return type because nobody cares when its done: this is method for freeing stream related resources during a session panic. I should make the docs more clear about that, and maybe the name as well.

Technically, it could be arbitrarily slow since the close will propagate down to stream pipelines which can then perform arbitrary work and may not be forking their Future's into a different thread. I consider that to be the responsibility of the user to use an `ExecutionContent` as appropriate, and shouldn't be a big deal.

http/src/main/scala/org/http4s/blaze/http/http2/StreamState.scala
bryce-anderson: Roughly the same answer: I'm need this to complete before I move on to maintain coherence of the session state. Eg, if I close a stream, I need to know that it is gone *now*.

http/src/main/scala/org/http4s/blaze/http/http2/StreamState.scala
bryce-anderson: Great idea, I hadn't considered that.

PR #137, opened 2017-12-29
Refine the notion of the StreamManager

bryce-anderson:
The StreamManager now takes on a larger role of ensuring a constant session 
state with regard to which streams are idle, closed, open, etc.

Also fixed some bugs in SessionFlowControlImpl where the wrong window
was being used in checks.


bryce-anderson:
I'd like to get this merged, but I understand that reviewers may be a bit busy with other things at this time. I'd like to get this merged by the end of the week and folks can review post-facto. Any objections to that?

bryce-anderson:
@ChristopherDavenport for testability, mostly, since traits are easier to mock. I've also gotten used to the pattern of separating interface from implementation, perhaps to much so.

PR #138, opened 2017-12-29
Add the SessionFrameListener

bryce-anderson:
Implementation of a FrameListener that operates on the state of an
underlying SessionCore.


bryce-anderson:
Note that this is dependant on #137 and will be rebased on that commit once it's merged.

PR #139, opened 2018-01-06
Add the SessionFrameListener

bryce-anderson:
Implementation of a FrameListener that operates on the state of an
underlying SessionCore.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/Http2Settings.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/Http2Settings.scala
@@ -154,12 +154,12 @@ object Http2Settings {
       else None
   }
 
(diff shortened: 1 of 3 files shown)


bryce-anderson:
In the essence of keeping things moving along, I'm going to merge this sometime before the end of the week barring objections. 

After this the only things left to do are merge the handshake stuff and the concrete session impls (which are pretty thin: mostly just concrete implementations of the `SessionCore`)! Oh, and tests. Lots of tests.

rossabaker:
I'm not feeling well.  I'll try, but it probably won't be tonight.

bryce-anderson:
@ChristopherDavenport ty. Embedding spec is one of the few times that I actually enjoy adding comments to code since the spec is not going to go stale anytime soon.

@rossabaker, no rush: I don't expect to work off of this until the weekend. And feel better.

PR #142, opened 2018-01-14
Remote Http2ClientConnection and cleanup some interfaces

bryce-anderson:
We can get by just fine with Http2ClientSession if we add a `ping()`
method.


bryce-anderson:
Thanks for the review @ChristopherDavenport!

PR #144, opened 2018-01-15
Cleanup of unused and/or redundant abstractions

bryce-anderson:
Some general cleanup work including but not limited to:
* Removing the unused `Halt` object
* Defining `isClient` for the SessionCore in terms of having a builder
  for inbound stream.


http/src/main/scala/org/http4s/blaze/http/http2/StreamManagerImpl.scala
```diff
@@ -102,18 +104,17 @@ private final class StreamManagerImpl(
...
-        val streamState = new InboundStreamStateImpl(session, streamId, streamFlowWindow)
-        assert(streams.put(streamId, streamState).isEmpty)
-
-        // assemble the pipeline
-        leafBuilder.base(streamState)
-        streamState.sendInboundCommand(Command.Connected)
-
-        Right(streamState)
+    } else {
+      val leafBuilder = inboundStreamBuilder.get.apply(streamId) // we alread made sure it wasn't empty
```
rossabaker: *already

PR #145, opened 2018-01-20
Introduce ConnectionImpl

bryce-anderson:
This is a concrete implementation of a HTTP connection state machine.


bryce-anderson:
I'm probably going to merge this sometime in the afternoon in order to keep this rolling: still plenty of code left before it's good to go.

aeons:
I took a look at this, but I'm far from comfortable enough with blaze internals to add anything valuable, I think :)

bryce-anderson:
@aeons, thanks for taking a look: it's always better than nothing.

PR #146, opened 2018-01-21
Refactor Settings related parts of codecs and listeners

bryce-anderson:
Since an ack and settings are mutually exclusive, we can model it
better as Settings(Option[Seq[Setting]]) where None represents an ack
and Some represents a settings frame.


bryce-anderson:
Since this is pretty much mechanical work I'm going to merge it pretty quick, presuming Travis gives me the green light.

rossabaker:
:+1:

PR #147, opened 2018-01-21
Add PriorKnowledgeHandshaker for client and server

bryce-anderson:
They are in change of handling the connection preface and sending and
receiving the initial settings frames.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/PriorKnowledgeHandshaker.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/PriorKnowledgeHandshaker.scala
@@ -1,7 +1,6 @@
 package org.http4s.blaze.http.http2
 
 import java.nio.ByteBuffer
-import java.nio.charset.StandardCharsets
 
 import org.http4s.blaze.http.http2.SettingsDecoder.SettingsFrame
 import org.http4s.blaze.pipeline.{Command, TailStage}
@@ -39,7 +38,7 @@ abstract class PriorKnowledgeHandshaker[T](localSettings: ImmutableHttp2Settings
   }
 
   private[this] def handleSettings(bytes: ByteBuffer): Future[(MutableHttp2Settings, ByteBuffer)] = {
-    sendSettings().flatMap { _ => receiveSettings(bytes) }
+    sendSettings().flatMap { _ => readSettings(bytes) }
   }
 
   private[this] def sendSettings(): Future[Unit] = {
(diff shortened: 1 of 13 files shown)


http/src/main/scala/org/http4s/blaze/http/http2/PriorKnowledgeHandshaker.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  private[this] def sendSettings(): Future[Unit] = {
+    val settingsBuffer = FrameSerializer.mkSettingsFrame(localSettings.toSeq)
+    channelWrite(settingsBuffer)
+  }
+
+  private[this] def receiveSettings(acc: ByteBuffer): Future[(MutableHttp2Settings, ByteBuffer)] = {
+    logger.debug("receiving settings")
+    getFrameSize(acc) match {
+      case -1 =>
```
ChristopherDavenport: This is without a settings frame then? Handles getting information till we finally have one? Not sure I follow this section.

http/src/main/scala/org/http4s/blaze/http/http2/client/ClientSessionImpl.scala
```diff
@@ -0,0 +1,35 @@
...
+
+private final class ClientSessionImpl(
+    tailStage: TailStage[ByteBuffer],
+    localSettings: ImmutableHttp2Settings,
+    remoteSettings: MutableHttp2Settings,
+    flowStrategy: FlowStrategy,
+    parentExecutor: ExecutionContext)
+  extends Http2ClientSession {
+
+  override def dispatch(request: HttpRequest): Future[ReleaseableResponse] = ???
```
ChristopherDavenport: This is for a future PR I assume.

http/src/main/scala/org/http4s/blaze/http/http2/server/ServerPriorKnowledgeHandshaker.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import org.http4s.blaze.http.http2._
+import org.http4s.blaze.pipeline.stages.{BasicTail, OneMessageStage}
+import org.http4s.blaze.pipeline.{Command, LeafBuilder}
+import org.http4s.blaze.util.{BufferTools, Execution, StageTools}
+
+import scala.concurrent.Future
+import scala.util.Failure
+
+// TODO: can we reduce the visibility of this?
```
ChristopherDavenport: Reduce Visibility How? Could Embed Inside an Object of Some Sort.

http/src/test/scala/org/http4s/blaze/http/http2/ProtocolFrame.scala
```diff
@@ -8,7 +8,7 @@ sealed abstract class ProtocolFrame private extends Product with Serializable
 object ProtocolFrame {
   case class GoAway(lastHandleStream: Int, cause: Http2Exception) extends ProtocolFrame
 
-  case class Ping(data: Option[Array[Byte]]) extends ProtocolFrame
+  case class Ping(isAck: Boolean, data: Array[Byte]) extends ProtocolFrame
```
ChristopherDavenport: Cleaner, are Ping and Ack really the same Frame? That seems odd, but the new repr is cleaner.

http/src/test/scala/org/http4s/blaze/http/http2/ServerPriorKnowledgeHandshakerSpec.scala
```diff
@@ -0,0 +1,116 @@
...
+      val head = new MockHeadStage
+      val handshaker = makeHandshaker(localSettings)
+      LeafBuilder(handshaker).base(head)
+
+      val serverSession = handshaker.handshake()
+      head.reads.dequeue().success(bits.getPrefaceBuffer())
+
+      // Consume all the inbound data which may be a prelude and the local settings
+      head.consumeOutboundData()
+      head.consumeOutboundData()
```
ChristopherDavenport: This is called twice for prelude and local settings separately?

http/src/main/scala/org/http4s/blaze/http/http2/PriorKnowledgeHandshaker.scala
bryce-anderson: Correct: if we don't have enough data yet, we read again, concat the data, and try again.

http/src/main/scala/org/http4s/blaze/http/http2/client/ClientSessionImpl.scala
bryce-anderson: Yes, just a stub. Sorry, I'll add a comment to that effect.

http/src/main/scala/org/http4s/blaze/http/http2/server/ServerPriorKnowledgeHandshaker.scala
bryce-anderson: I was thinking of just the package: maybe it doesn't belong in `http` but instead `server` or `http2`. But thinking about it more, I don't think it really matters so long as it doesn't escape the blaze scope: at that point we can be free to change it any time.

http/src/test/scala/org/http4s/blaze/http/http2/ProtocolFrame.scala
bryce-anderson: They are exact mirrors with the exception of the ACK flag, this previously it was actually not a true representation. The 8 bytes are typically a stamp of some sort and may have meaning to the client, for example, to disentangle multiple pings, although the server _should_ send acks in the same order as pings are received.

http/src/test/scala/org/http4s/blaze/http/http2/ServerPriorKnowledgeHandshakerSpec.scala
bryce-anderson: Correct. I'll add a note to make that clear because it does look suspect.

http/src/main/scala/org/http4s/blaze/http/http2/PriorKnowledgeHandshaker.scala
bryce-anderson: I think I wasn't clear: -1 means that there wasn't even enough data for the frame header so a size couldn't be determined. I've changed the result of `getFrameSize` to an `Option[Int]` to better reflect that. Thanks again for pointing that out.

PR #148, opened 2018-01-30
Add ClientStage and ClientSessionImpl

bryce-anderson:
The ClientStage is responsible for handling the dispatch of a HTTP
request which means turning a request and response into and from a
stream of HTTP/2 messages.

The ClientSessionImpl is just a thin wrapper around the HttpConnection.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/client/ClientStage.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/client/ClientStage.scala
@@ -1,6 +1,7 @@
 package org.http4s.blaze.http.http2.client
 
 import java.nio.ByteBuffer
+import java.util.Locale
 
 import org.http4s.blaze.http.{ClientResponse, _}
 import org.http4s.blaze.http.HttpClientSession.ReleaseableResponse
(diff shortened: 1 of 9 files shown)


bryce-anderson:
After this all we need is the `Http2ClientSelector` (probably renamed), and a basic session manager in order to get the first H2 client merged!

http/src/main/scala/org/http4s/blaze/http/http2/client/ClientStage.scala
```diff
@@ -0,0 +1,198 @@
...
+  }
+}
+
+private object ClientStage {
+  private[client] def makeHeaders(request: HttpRequest): Try[Vector[(String, String)]] = {
+    UrlComposition(request.url).map { breakdown =>
+      val hs = new VectorBuilder[(String, String)]
+
+      // h2 pseudo headers
+      hs += StageTools.Method -> request.method.toUpperCase
```
rossabaker: I guess we've ascertained that method doesn't include funny characters, but I still twitch when I see this without a `Locale.ROOT`.

bryce-anderson: We haven't checked the characters, so I'll add a sane `Locale`.

bryce-anderson: `Local.ENGLISH` it is. Thanks for the keen eye.

rossabaker: Somewhere along the way, I picked up `Locale.ROOT` as "the neutral locale", but I can't articulate any difference between the two.  The important thing here is that this isn't broken in Turkey.

PR #149, opened 2018-01-31
Add Http2Client and the session manager and selectors necessary

bryce-anderson:
Finally, we have a HTTP/2 client. It still needs a battery of
end-to-end tests which are run against a local server.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
@@ -17,6 +17,9 @@ import scala.util.control.NonFatal
 import scala.util.{Failure, Success}
 
 private object ClientSessionManagerImpl {
+
+  private val LowQualityThreshold = 0.1
+
   case class ConnectionId(scheme: String, authority: String)
 
   case class Http1SessionProxy(id: ConnectionId, parent: Http1ClientSession) extends Http1ClientSession {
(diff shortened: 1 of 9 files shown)


bryce-anderson:
I think the session managers as a whole need a large amount of work. Specifically, giving some thought to timeouts and failures would be nice.

http/src/main/scala/org/http4s/blaze/http/ClientSessionManagerImpl.scala
```diff
@@ -70,7 +70,7 @@ private final class ClientSessionManagerImpl(sessionCache: java.util.Map[Connect
             h2.closeNow() // make sure its closed
             it.remove()
 
-          case h2: Http2ClientSession if h2.quality > 0.1 =>
+          case h2: Http2ClientSession if 0.1 < h2.quality =>
```
rossabaker: What prompted this flip?

bryce-anderson: Just that I think it reads easier.

jcranky: Maybe while at it, `0.1` could also get a name and be even easier to read?

bryce-anderson: Done.

bryce-anderson:
I think this is about ready, more feedback appreciated. That said, no rush: the server side is what is left and is kind of orthogonal to this.

PR #150, opened 2018-02-07
Add the server side RequestParser

bryce-anderson:
The RequestParser turns a sequence of pseudo-headers and a body into a
legit HttpRequest while rejecting malformed request preludes according
to the spec.

Also some smaller cleanups included.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/server/RequestParser.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/server/RequestParser.scala
@@ -1,0 +1,80 @@
+package org.http4s.blaze.http.http2.server
+
+import org.http4s.blaze.http.http2.StageTools._
+import org.http4s.blaze.http.{BodyReader, HttpRequest, _}
+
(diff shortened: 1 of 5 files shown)


http/src/main/scala/org/http4s/blaze/http/http2/server/RequestParser.scala
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+      case h@(k, v) => // Non pseudo headers
+        pseudoDone = true
+        k match {
+          case Connection =>
+            error += s"HTTP/2.0 forbids connection specific headers: $h. "
+
+          case TE =>
+            if (!v.equalsIgnoreCase("trailers")) error += s"HTTP/2.0 forbids TE header values other than 'trailers'. "
```
ChristopherDavenport: Is the interpolator necessary here?

bryce-anderson: Nope, I'll remove it.

PR #151, opened 2018-02-08
Add the server AbstractBodyReader

bryce-anderson:
It's role is to provide the application space a handle on the data from
the HTTP/2 stream.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
@@ -1,0 +1,90 @@
+package org.http4s.blaze.http.http2.server
+
+import java.nio.ByteBuffer
+
+import org.http4s.blaze.http.BodyReader
+import org.http4s.blaze.http.http2.{DataFrame, HeadersFrame, StreamMessage}
+import org.http4s.blaze.http.http2.Http2Exception._
+import org.http4s.blaze.util.{BufferTools, Execution}
+import org.log4s.getLogger
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+// TODO: this could probably be shared with the client
+/** Base implementation of the HTTP/2 [[BodyReader]]
+  *
+  * @param streamId the HTTP/2 stream id associated with this reader
+  * @param length length of data expected, or -1 if unknown.
+  */
+private abstract class AbstractBodyReader(streamId: Int, length: Long) extends BodyReader {
+  import AbstractBodyReader._
+
+  private[this] var bytesRead = 0L
+  private[this] var finished = false
+
+  private[this] val lock: Object = this
+  private[this] val logger = getLogger
+
+  protected def channelRead(): Future[StreamMessage]
+
+  protected def failed(ex: Throwable): Unit
+
+  override def discard(): Unit = lock.synchronized {
+    finished = true
+  }
+
+  override def isExhausted: Boolean = lock.synchronized(finished)
+
+  def apply(): Future[ByteBuffer] = {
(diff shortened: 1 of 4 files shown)


http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
```diff
@@ -0,0 +1,90 @@
...
+              // optionally, one HEADERS frame, followed by zero or more
+              // CONTINUATION frames containing the trailer-part, if present (see
+              // [RFC7230], Section 4.1.2).
+              // https://tools.ietf.org/html/rfc7540#section-8.1
+              finished = true
+              val msg = "Received non-final HEADERS frame while reading body."
+              Failure(PROTOCOL_ERROR.rst(streamId, msg))
+          }
+        }
+        // TODO: change to the `.fromTry` API once we've dumped 2.10.x builds
```
ChristopherDavenport: Since we're not supporting 2.10 in http4s the new code likely can dump 2.10 if you would like.

http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
```diff
@@ -0,0 +1,90 @@
...
+  * @param streamId the HTTP/2 stream id associated with this reader
+  * @param length length of data expected, or -1 if unknown.
+  */
+private abstract class AbstractBodyReader(streamId: Int, length: Long) extends BodyReader {
+  import AbstractBodyReader._
+
+  private[this] var bytesRead = 0L
+  private[this] var finished = false
+
+  private[this] val lock: Object = this
```
ChristopherDavenport: How does lock.synchronized over the `this` reference work?

http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
```diff
@@ -0,0 +1,90 @@
...
+              }
+
+            case HeadersFrame(_, true, ts) =>
+              finished = true
+              logger.info(s"Discarding trailer headers: $ts")
+              Success(BufferTools.emptyBuffer)
+
+            case HeadersFrame(_, false, _) =>
+              // trailers must be the final frame of the stream:
+              // optionally, one HEADERS frame, followed by zero or more
```
ChristopherDavenport: I'm not sure I understand how this line is accounted for here.

http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
bryce-anderson: I'd love to, I was just hoping not to make it part of this diff. But, alas, this is as good a breaking point as any, so I'll probably do it.

http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
bryce-anderson: It basically just makes everything synchronize on `this`. I don't know where I developed this pattern, but I typically like to have an explicit object that is being locked on, but perhaps this is just added noise. I'll just use the more idiomatic ` synchronized {`.

http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyReader.scala
bryce-anderson: I suspect you mean the comment. It's an excerpt of a much larger paragraph. I'll either make that more clear, or just insert the whole thing. Sorry for the confusion.

bryce-anderson:
Thanks for the review @ChristopherDavenport!

PR #153, opened 2018-02-10
Add the HTTP/2 ServerStage

bryce-anderson:
It drives the dispatch of individual stream requests and is a
one-time use abstraction: one per stream.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http2/server/ServerStage.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/server/ServerStage.scala
@@ -1,0 +1,148 @@
+package org.http4s.blaze.http.http2.server
+
+import java.nio.ByteBuffer
+import java.util.Locale
+
+import org.http4s.blaze.http._
+import org.http4s.blaze.http.http2.Http2Exception._
+import org.http4s.blaze.http.http2.StageTools._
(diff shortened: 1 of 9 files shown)


http/src/main/scala/org/http4s/blaze/http/http2/server/ServerStage.scala
```diff
@@ -0,0 +1,144 @@
...
+        timeoutService(request).onComplete(renderResponse(request.method, _))(config.serviceExecutor)
+
+      case Left(errMsg) =>
+        shutdownWithCommand(Cmd.Error(PROTOCOL_ERROR.rst(streamId, errMsg)))
+    }
+  }
+
+  private[this] def renderResponse(method: String, response: Try[RouteAction]): Unit = response match {
+    case Success(builder) =>
+      builder.handle(getWriter(method == "HEAD", _))
```
ChristopherDavenport: Its clear in the context of the tests, but not clear at the call that this == is the check that is forcing the NoOp writer on a head request. Once you go to the other functions its more understandable, but it would nice to be clearer at the invocation site.

bryce-anderson: Thats a fair point, and I vaguely remember thinking it was a bit tricky at some point as well. I'll make the code a hair more verbose and/or add some comments to clear it up.

PR #154, opened 2018-02-10
Add the HTTP/2 ServerSelector

bryce-anderson:
The final piece needed for making HTTP/2 server sessions

### DIFF
--- a/examples/src/main/scala/org/http4s/blaze/examples/http2/Http2ServerExample.scala
+++ b/examples/src/main/scala/org/http4s/blaze/examples/http2/Http2ServerExample.scala
@@ -1,0 +1,37 @@
+package org.http4s.blaze.examples.http2
+
+import java.net.InetSocketAddress
+
+import org.http4s.blaze.channel
+import org.http4s.blaze.channel._
+import org.http4s.blaze.channel.nio1.NIO1SocketServerGroup
+import org.http4s.blaze.examples.{ExampleKeystore, ExampleService}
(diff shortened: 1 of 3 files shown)


bryce-anderson:
Note that this depends upon #153.

bryce-anderson:
Also, this will likely be the last commit before I go about merging the `merge-new-client-http2` branch into `master`. :)

examples/src/main/scala/org/http4s/blaze/examples/http2/Http2ServerExample.scala
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  private val factory = NIO1SocketServerGroup.fixedGroup(workerThreads = channel.DefaultPoolSize)
+
+  def run(): ServerChannel = factory.bind(new InetSocketAddress(port), f).getOrElse(sys.error("Failed to start server."))
+}
+
+object Http2ServerExample {
+  def main(args: Array[String]): Unit = {
+    println("Hello world!")
```
ChristopherDavenport: Whats this `println` about? Just to signify execution right before the server should start?

bryce-anderson: Yeah, and it should be something else more meaningful. Thanks for pointing it out.

bryce-anderson:
@ChristopherDavenport, don't flatter me. 😉  There shouldn't be anything particularly tricky here, so if it was clear then that is a great sign.

PR #160, opened 2018-02-11
Improve the end-to-end tests

### LINKED ISSUE #158: Add end-to-end tests for HTTP/2
We don't have much in the way of end-to-end tests for the HTTP/2 implementation. Part of the problem is that it's tricky to add in unit-test form because we need the have the Jetty ALPN boot classes on the classpath.

I think that Jetty supports the direct upgrade, so that could help.

bryce-anderson:
Closes #158

We currently have a few tests, but they are difficult to maintain. This
introduces a scaffold structure so we can write tests generically and
run them against different combinations of implementations.


http/src/test/scala/org/http4s/blaze/http/endtoend/scaffolds/ClientScaffold.scala
```diff
@@ -0,0 +1,30 @@
...
+  // Don't be afraid to block: this is for testing.
+  def runRequest(request: HttpRequest): Response
+
+  final def runGet(url: String, headers: Headers = Nil): Response = {
+    val request = HttpRequest("GET", url, majorVersion, minorVersion, headers, BodyReader.EmptyBodyReader)
+    runRequest(request)
+  }
+
+  final def runPost(url: String, bodyReader: BodyReader, headers: Headers = Nil): Response = {
+    val request = HttpRequest("POST", url, 1, 1, headers, bodyReader)
```
ChristopherDavenport: Major, Minor Here?

bryce-anderson: Yup, this is a mistake. Thanks for the keen eye!

http/src/test/scala/org/http4s/blaze/http/endtoend/scaffolds/AsyncHttpClientScaffold.scala
```diff
@@ -0,0 +1,68 @@
...
+
+import scala.collection.JavaConverters._
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+/** Basic scaffold for running http requests against a server
+ *
+  * @param timeout timeout for request. Must be finite and greater than 0
+  */
+private[endtoend] final class AsyncHttpClientScaffold(timeout: Duration) extends ClientScaffold(1, 1) {
```
ChristopherDavenport: Why AsyncHttpClient instead of Http1ClientScaffold in line with the other naming conventions in the scaffolds?

bryce-anderson: The name does suck, though I'd like to keep the name broadcasting that it's the `async-http-client` since I'd like to have a couple implementations at the end of the day.

PR #165, opened 2018-02-12
SelectorLoop cleanup

### LINKED ISSUE #164: Tasks can be lost in the SelectorLoop
Tasks can get lost on the [`SelectorLoop`](https://github.com/http4s/blaze/blob/bb56a6b3238685a820a72e19a20ea23cce26cbb2/core/src/main/scala/org/http4s/blaze/channel/nio1/SelectorLoop.scala#L51) due to a race condition between closing and the loop shutting down.

bryce-anderson:
Fixes #164 

Cleanup the `SelectorLoop`. While doing so a task leak was discovered where a `Runnable` could be added to the `SelectorLoop`s queue of tasks to execute, but would never be executed if the selector got closed at a slightly unlucky time.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
@@ -22,10 +22,11 @@ private[nio1] object NIO1HeadStage {
 
 }
 
-private[nio1] abstract class NIO1HeadStage(ch: SelectableChannel,
-                                         loop: SelectorLoop,
-                                          key: SelectionKey) extends ChannelHead
-{
+private[nio1] abstract class NIO1HeadStage(
+  ch: SelectableChannel,
+  loop: SelectorLoop,
+  key: SelectionKey
+) extends ChannelHead {
   import NIO1HeadStage._
 
   override def name: String = "NIO1 ByteBuffer Head Stage"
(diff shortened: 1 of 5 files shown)


bryce-anderson:
Yes, though I don't think it's a serious one: in practice I don't think sector loops are often closed and all the work submitted is typically the business of members of the loop which should be closed anyway.

core/src/main/scala/org/http4s/blaze/channel/nio1/SelectorLoop.scala
```diff
@@ -5,165 +5,137 @@ import org.http4s.blaze.pipeline.{Command => Cmd}
...
-      runTasks()
+      // Run any pending tasks. These may set interest ops,
+      // just compute something, etc.
+      taskQueue.executeTasks()
 
-    // Block here until some IO event happens or someone adds a task to run and wakes the loop
-      if (selector.select() > 0) {
+      // Block here until some IO event happens or someone adds
+      // a task to run and wakes the loop
+      if (0 < selector.select()) {
```
hvesalai: `selector.select() > 0` would be nicer to read.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
```diff
@@ -125,6 +102,34 @@ private[nio1] abstract class NIO1HeadStage(ch: SelectableChannel,
     }
   }
 
+  ///  channel reading bits //////////////////////////////////////////////
+
+  final override def readRequest(size: Int): Future[ByteBuffer] = {
```
hvesalai: Size is not used (other than for logging. For readability it would be better to not take such parameter as now the signature of the method gives the impression that some logic is contained to attempt reading `size` bytes.

bryce-anderson: That was the original idea of the `HeadStage.read(Int)` method, but as it turns out, it's not all that useful. I'd love to remove it, and I just filed a ticket for it: https://github.com/http4s/blaze/issues/166.

bryce-anderson:
Barring objections, I'm going to merge this sometime this evening.

PR #169, opened 2018-02-14
Rename StreamMessage to StreamFrame

bryce-anderson:
This naming is more consistent with what it represents and the
subclasses DataFrame and HeadersFrame.


bryce-anderson:
Since this is search-and-replace, I'm probably going to merge it once I get the all-clear from CI.

PR #170, opened 2018-02-14
Fix locking issue in Http1ServerCodec

bryce-anderson:
The Http1ServerCodec was satisfying a Promise while holding a lock which
is generally poor form since it can result in arbitrary execution and
potentially deadlocks.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
@@ -138,26 +138,26 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
     }
   }
 
-  // Must be called from within a `lock.synchronized` block
   private[this] def readAndGetRequest(p: Promise[HttpRequest]): Unit = {
     // we need to get more data
     pipeline.channelRead().onComplete {
       case Success(buff) =>
-        lock.synchronized {
-          try {
+        try {
+          val httpRequest = lock.synchronized {
             buffered = BufferTools.concatBuffers(buffered, buff)
-            val httpRequest = maybeGetRequest()
-            if (httpRequest != null) {
-              p.trySuccess(httpRequest)
-            }
-            else {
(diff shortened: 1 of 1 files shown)


rossabaker:
Why is the Java reformatting here?

LGTM. Is this severe enough to backporr to 0.12?

bryce-anderson:
I'll break it up into two separate commits. The bug only affects those using the raw blaze server since 0.13.0, so 0.12 is in the clear.

rossabaker:
Is 0.13 a maintained version or EOL? I've tried to merge 0.12 through it on the way to 0.14, but never bothered to release.  At this point, I'm for putting resources into 0.14 unless something is broken for http4s users on 0.12.

bryce-anderson:
I'm considering 0.13 EOL and putting everything into 0.14.x outside of serious bug fixes for 0.12.x, which don't seem all that common.

I've got a local branch of http4s proper I'm working on to get it onto the 0.14.x series. Right now there are a few things I've had to pitch over the wall (websockets stuff and `Serializers`) and I need to expand the visibility of some http2 stuff. I'm hoping to get a branch working and then do a blaze 0.14.0-M2 in a week or two, presuming I can find the time.

bryce-anderson:
@ChristopherDavenport I don't have a link for your, but this has resulted in deadlocks at my day job. I can offer a generic example:

```scala
val, p1, p2 = Promise[Unit]
val lock1, lock2 = someLocks
p1.flatMap{ _ => lock2.synchronized { .. usually the is hidden in a method .. } }(ThreadLocalExecutor)

p2.flatMap{ _ => lock1.synchronized { .. usually the is hidden in a method .. } }(ThreadLocalExecutor)

Thread 1:
lock1.synchronized( p1.success(())) // pauses right before taking lock2

Thread2:
lock2.synchronized(p2.success(())) // gets to the point where it's waiting on lock1, held by Thread1

Thread1 continues, but can't take lock2 since it's held by Thread2. Thread1 and Thread2 are now deadlocked.
```
This is exactly what happened at my day job, resulting in the deadlocking of event loop threads, which was obviously a no-go.

PR #173, opened 2018-02-15
Start opening up the HTTP/2 types for http4s proper consumption

bryce-anderson:
We made a lot of the API private, but some of it should be made public
again for consumption by http4s proper.

### DIFF
--- a/examples/src/main/scala/org/http4s/blaze/examples/ExampleService.scala
+++ b/examples/src/main/scala/org/http4s/blaze/examples/ExampleService.scala
@@ -5,7 +5,6 @@ import java.nio.charset.StandardCharsets
 import java.util.concurrent.atomic.AtomicReference
 
 import org.http4s.blaze.channel.ServerChannel
-import org.http4s.blaze.http.util.HeaderNames
 import org.http4s.blaze.http._
 import org.http4s.blaze.pipeline.stages.monitors.IntervalConnectionMonitor
 import org.http4s.blaze.util.{BufferTools, Execution}
--- a/http/src/main/scala/org/http4s/blaze/http/http2/server/RequestParser.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http2/server/RequestParser.scala
@@ -1,14 +1,15 @@
 package org.http4s.blaze.http.http2.server
 
-import org.http4s.blaze.http.http2.StageTools._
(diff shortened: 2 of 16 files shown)


bryce-anderson:
@ChristopherDavenport, that's a good point. The real value is in the pseudo-header names. I'll move them into a separate object that is public and leave `StageTools` the way it is.

bryce-anderson:
@ChristopherDavenport, I think this is better now if you care to take a second look.

examples/src/main/scala/org/http4s/blaze/examples/ExampleService.scala
```diff
@@ -5,8 +5,7 @@ import java.nio.charset.StandardCharsets
 import java.util.concurrent.atomic.AtomicReference
 
 import org.http4s.blaze.channel.ServerChannel
-import org.http4s.blaze.http.util.HeaderNames
-import org.http4s.blaze.http._
+import org.http4s.blaze.http.{HeaderNames, _}
```
rossabaker: Uh, doesn't the underscore cover this?

http/src/main/scala/org/http4s/blaze/http/http2/server/RequestParser.scala
```diff
@@ -1,12 +1,14 @@
...
 
-import org.http4s.blaze.http.http2.StageTools._
 import org.http4s.blaze.http.{BodyReader, HttpRequest, _}
 
 import scala.collection.mutable.ArrayBuffer
 
 /** Tool for turning a HEADERS frame into a request */
 private object RequestParser {
+  import org.http4s.blaze.http.http2.PseudoHeaders._
+  import org.http4s.blaze.http.http2.StageTools._
```
rossabaker: Any reason not to put these with the rest of the imports?

examples/src/main/scala/org/http4s/blaze/examples/ExampleService.scala
bryce-anderson: Yeah. IntelliJ rename fail.

http/src/main/scala/org/http4s/blaze/http/http2/server/RequestParser.scala
bryce-anderson: Personal preference I suppose. I'll move it.

PR #175, opened 2018-02-16
Add scalastyle to the build and fix some violations

### LINKED ISSUE #174: Add scalastyle to the build
Now that everything is in master it might be a good time to conform to some sort of style.

bryce-anderson:
Closes #174

This gets us on the track to uniformity.


bryce-anderson:
Looks like I poorly named the branch...

bryce-anderson:
There are definitely a number of unheeded warnings, most around magic numbers (places like [here](https://github.com/http4s/blaze/blob/master/http/src/main/scala/org/http4s/blaze/http/http2/Http2Exception.scala#L100) where I don't think they should be considered magic) and a couple places about complexity that didn't seem right to address in this PR considering the risk.

Since this is a big step in the right direction I'll merge this and we can continue to iterate in the future.

PR #177, opened 2018-02-16
Add scalafmt to the project and apply it

bryce-anderson:
Closes #175

All changes where performed by `scalafmt` other than those to the plugins.sbt and BlazePlugin.scala files.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,14 @@
+style = default
+
+maxColumn = 100
+
+// Vertical alignment is pretty, but leads to bigger diffs
+align = none
+
+rewrite.rules = [
+  AvoidInfix
+  RedundantBraces
+  RedundantParens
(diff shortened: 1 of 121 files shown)


rossabaker:
[Slightly](https://github.com/http4s/http4s/blob/70870ae9d38449a2ae3f1207ad72085a7af2a699/.scalafmt.conf). `align = none` is the one I'd strongly advocate for, as it reduces diff noise.  The rest are more optional, though I guess it'd be nice to be consistent through the org absent a strong reason not to.

bryce-anderson:
Ah, nice. That is much better. And the line lengths are a little more lenient with the http4s config.

bryce-anderson:
I just rebased onto master now that it's dependent commit was merged to master. It applied cleanly considering there were precisely no changes to code required.

PR #178, opened 2018-02-16
Move the validateHeaderName to the public HeaderNames object

bryce-anderson:
Ths method is also useful for http4s.


bryce-anderson:
It turns out that I also need this from `StageTools` (or to duplicate it). It was just a move/rename, no behavior changes.

PR #185, opened 2018-02-18
NIO1 event loops use a ThreadFactory to spawn threads

### LINKED ISSUE #183: Start using ThreadFactories for the nio1 selector things
It would be nice if the user could configure the threads responsible for the selector loops, although they should still have a sane default.

bryce-anderson:
Closes #183.

Using a ThreadFactory to generate the threads for the SelectorLoops
and NIO1SocketServerGroup gives users an easy way to control how
threads are created and their properties.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/ServerChannelGroup.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/ServerChannelGroup.scala
@@ -7,6 +7,11 @@ import org.log4s.getLogger
 
 import scala.util.Try
 
+/** Abstraction for binding a server socket and handling connections.
+  *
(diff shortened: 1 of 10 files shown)


bryce-anderson:
The `NIO1SocketServerGroup` looks like a massive change but it's not actually a lot of change other than indent, which was poorly captured by GitHub.

bryce-anderson:
I'm not thrilled about having two different `ThreadFactory`s: one for the accept look and one for the workers. I've opened #186 to clean that up, but I'd like to do it after this is merged since they will clearly conflict.

bryce-anderson:
This has been rebased on master. Although the changes to `ChannelSpec` look large, they are strictly cosmetic.

PR #187, opened 2018-02-19
Make the SelectorLoop listen for generic events

### LINKED ISSUE #186: SelectorLoops should select more generally
Right now they are specific to registering `NIO1HeadStage` objects, but that excludes listening for inbound or outbound connections. It should be made more general, maybe an interface like
```scala
trait Selectable {
  def opsReady(scratch: ByteBuffer): Unit
  def close(): Unit
}
```
That should give the general handles for notifying when ops are ready and a handle for the loop to force close resources.

That will allow us to have the server listen using any old standard loop so it won't need it's own thread which makes #183 so ugly.

bryce-anderson:
Closes #186.

The `SelectorLoop` is hard coded to the `NIO1HeadStage` type which means it's really only useful for selection I/O events on established sockets. By abstracting the type registered with the `SelectorLoop` to the `Selectable` trait, we can listen to all events supported by `SelectableSocket`s. This allows us to drop the dedicated server socket listening thread and just register interest in those events with any `SelectorLoop`.


bryce-anderson:
I decided that I wanted this to come before #185 since it will avoid some transient (and bad) API changes. It shouldn't be terribly tough to redo the work especially now that there is one less `ThreadFactory` to consider.

bryce-anderson:
Sorry for all the big changes in pretty hairy places, though I think things are getting quite a bit nicer.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
```diff
@@ -214,27 +221,28 @@ private[nio1] abstract class NIO1HeadStage(
       }
 
       writeData = null
+      try ch.close()
+      catch {
+        case ex: IOException => logger.debug(ex)("During close")
```
rossabaker: debug?

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
```diff
@@ -49,30 +54,117 @@ object NIO1SocketServerGroup {
...
+        try acceptNewConnections()
+        catch {
+          case ex: IOException =>
+            closeWithError(ex)
+        }
+      }
+
+    override def close(): Unit =
+      if (closed.compareAndSet(false, true)) {
+        logger.info(s"Listening socket(${ch.socketAddress})")
```
rossabaker: This statement confuses me.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
```diff
@@ -49,30 +54,117 @@ object NIO1SocketServerGroup {
...
+      } catch {
+        case _: RejectedExecutionException =>
+          logger.info("Selector loop closed. Closing in local thread.")
+          doClose()
+      }
 
-  // Start the selector thread
-  t.setDaemon(true)
-  t.start()
+    // Must be called within the `SelectorLoop`, it at all possible.
```
rossabaker: if at all possible

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
bryce-anderson: Debug seemed appropriate considering we're trying to end the connection. Honestly I don't know why that method can throw at all: if it throws and won't close, there is no way to release the resource, and if it closes and throws, what was the point of throwing?

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1SocketServerGroup.scala
bryce-anderson: It's clearly missing something... I'll fix it.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
rossabaker: I'm concerned about masking an an underlying problem.  We've done all we can, and the caller can't do more, so I agree it's not a throw.  It shouldn't affect the user experience, so I agree it's not a `log.error`.  But something is wrong that I would want to understand, so I think it's a `log.warn`.

bryce-anderson: Fair enough, I'll make it a `warn`.

Issue #189, opened 2018-02-25
Try removing the `InboundCommand`s


bryce-anderson:
The current pipeline structure is pull-based with read and write operations returning futures, so in some ways, they are like a version of a socket that return a Future instead of blocking. 

Exception for the `InboundCommand`s: they provide a way for the pipeline to _push_ events, which is a little out of place in the model. That said, they are useful for things like starting up a pipeline in a server where the server itself is the driver.

I think we can do better though, maybe making signaling startup as part of pipeline builders duties or changing what the server uses to bootstrap a pipeline. For example, it seems reasonable to provide a `HeadStage[ByteBuffer]` to a new form:
```scala
type BufferPipelineBuilder = SocketConnection => LeafBuilder[ByteBuffer]
// becomes
type BufferPipelineBuilder = HeadStage[ByteBuffer] => Unit
```
We'd need to do something to incorporate the info in `SocketConnection` into `HeadStage` but that seems like a good idea as well.

bryce-anderson:
This is on hold, if not canceled. @hvesalai has a pretty good point about wanting to be able to propagate cancellation, and without a push-based mechanism, that isn't easy. Inbound commands are becoming advisory commands now, so that addresses some of the difficulty about reasoning. They may be rebranded as `InboundEvent`s instead of `InboundCommand`s to make that more clear.

PR #190, opened 2018-02-25
Remove the `Disconnected` command

bryce-anderson:
As part of the effort to remove the `InboundCommand`s from the Stage
architecture (#189), this removes the `Disconnected` command. The result
is that it is up to the user of the pipeline to free resources and is not
notified eagerly about detected disconnects. In practice, this isn't all
that useful since socket events are similarly lazy: I/O operations are
typically required to detect disconnects, so the pipelines ends up getting
a `Disconnected` command pushed to them and a `EOF` result in an
operation.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
@@ -10,8 +10,7 @@ import java.nio.channels.{CancelledKeyException, SelectableChannel, SelectionKey
 import java.util.concurrent.RejectedExecutionException
 
 import scala.concurrent.{Future, Promise}
-import scala.util.control.NonFatal
-import org.http4s.blaze.pipeline.Command.{Disconnected, EOF}
+import org.http4s.blaze.pipeline.Command.EOF
 import org.http4s.blaze.util.BufferTools
 
 private[nio1] object NIO1HeadStage {
@@ -65,18 +64,13 @@ private[nio1] abstract class NIO1HeadStage(
         val p = readPromise
         readPromise = null
         if (p != null) {
-          p.tryComplete(r)
-          ()
-        } else {
-          /* NOOP: was handled during an exception event */
+          p.complete(r)
           ()
         }
 
       case Failure(e) =>
         val ee = checkError(e)
-        sendInboundCommand(Disconnected)
         closeWithError(ee) // will complete the promise with the error
-        ()
     }
   }
 
@@ -106,7 +100,6 @@ private[nio1] abstract class NIO1HeadStage(
 
       case WriteError(t) =>
         unsetOp(SelectionKey.OP_WRITE)
-        sendInboundCommand(Disconnected)
         closeWithError(t)
     }
   }
@@ -145,7 +138,7 @@ private[nio1] abstract class NIO1HeadStage(
     selectorLoop.executeTask(new Runnable {
       override def run(): Unit =
         if (closedReason != null) {
-          p.tryFailure(closedReason)
+          p.failure(closedReason)
           ()
         } else if (writePromise == null) {
           val writes = data.toArray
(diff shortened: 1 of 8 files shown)


bryce-anderson:
I'd appreciate feedback on the general idea here. My hope is that I can get ride of the `InboundCommand`s altogether since they make the model really awkward by smashing a small amount of push-based API into what is predominantly a pull-based API. I think the real kicker will be to see how removing the `Connected` command goes. That might involve changing the server bootstrapping API's to really make it work, so things might take a detour first.

hvesalai:
The use case for push-based Disconnet could be cancelling long running jobs.

hvesalai:
With that being said, I think it would be important to have the disconnect and further develop it to really try to cancel all work (at least Monix supports that).

bryce-anderson:
@hvesalai, can you point me to an example of cancelling a monix `Task`? I'm trying to figure out how a disconnected interrupt would be injected. I do want to convert blaze over to the cats `IO` at some point, which may support cancelation soon (https://github.com/typelevel/cats-effect/pull/121), but it appears to be at the execution level. So in future where `TailStage` looks kind of like:
```scala
trait TailStage {
  def read(): IO[ByteBuffer]
  def write(data: ByteBuffer): IO[Unit]
  def onInboundCommand(command: InboundCommand): Unit // pipeline events can come here
  ...
}
```
we won't have hooks into the execution fiber except through the `IO`s, so if a process isn't actually reading or writing, both of which can/do propagate the EOF condition, there isn't a way to inject the events from `onInboundCommand`. I don't actually know if what I've just proposed accurate, this is a legit question. 😄

At an absolute minimum, I want to change the semantics of `Disconnected` to be more to the tune of `Discarded`: I think it's critical for the driver of the stage to be in charge of closing the pipeline and freeing resources so as to avoid the current ambiguity as to who's job it is to do so.

hvesalai:
https://monix.io/api/2.3/monix/execution/Cancelable.html is the API for cancelling things.

Some things implement it, such as:

https://monix.io/api/2.3/monix/execution/CancelableFuture.html

While others return a Cancelable object, such as subscript.

https://monix.io/docs/2x/reactive/consumer.html#cancel-the-stream-on-subscription

hvesalai:
```
val future: CancelableFuture[Int] = task.runAsync
```

alexandru:
Noticed these comments. Just to add some details ...

In Monix 2.3 building a cancelable `Task` happens via `Task.create`, but the act of cancellation itself is something that happens at the edge of the world so to speak, being an FFI detail, something that happens in impure code, but also in `Task`'s operations that create race conditions, which use cancellation under the cover for releasing resources.

So we are talking for example of `Task.gather`, which executes tasks in parallel, which in case one task fails it cancels the rest. Or of the `.timeout(duration)` which triggers an error if a certain duration has passed without the source being completed, cancelling the source when the error gets emitted. Or the obvious `Task.chooseFirstOf` (renamed to `Task.race` in the 3.x line).

Due to popular request, in the 3.x line (currently at M3, soon to release an M4), we also have a new `Fiber` interface that is popped whenever a `Task` operation starts a background process, `Fiber` being the thing that's attached to it, sort of like a pure `CancelableFuture`. And for example we now have `Task.start` / `Task.fork`. And those operations yield a `Task[Fiber[A]]` as a result. And a `Fiber` has a pure `cancel: Task[Unit]` operation on it.

bryce-anderson:
@alexandru, thanks for the input, I really appreciate it! (and I really admire your work, so I'm a bit flattered that you've chimed in!)

So, at least with the old school way, it seems that the interruption gets signaled at the end of the world, by whoever is in charge of managing the execution (that doesn't mean it can't be signaled from elsewhere, but it is their job to do make the decision, or at least expose that property).

For a server, this seems achievable: typically the user passes some work but it is actually the server which manages the execution of the code. For a client, that's not usually the case, so the use case seems a bit fragmented.

For the `Fiber` interface which exposes a `cancel: Task[Unit]`, it sounds like a refinement on the same concept with a pure wrapper, is that correct?

Perhaps I need to quit writing code for a bit and spend some more time reading about the `Task` and `IO` ecosystem.

bryce-anderson:
I'm abandoning this. Instead of removing `Disconnected` and eventually `InboundCommand` they will be changed into advisory events. The first step is https://github.com/http4s/blaze/pull/192.

PR #191, opened 2018-02-27
ServerChannelGroup.acceptConnection is moved to `BufferedPipelinebuilder`

### LINKED ISSUE #184: ServerChannelGroup.acceptConnection is an idea that lives to deep
That makes rejecting connections an abstraction that lives all the way down at the `ServerChannelGroup`, which is silly. It really should be a part of the `BufferedPipelineBuilder`:
```scala
type BufferPipelineBuilder = SocketConnection => LeafBuilder[ByteBuffer]
// should be
type BufferPipelineBuilder = SocketConnection => Option[LeafBuilder[ByteBuffer]]
```
which is much easier for users.

bryce-anderson:
Closes #184.

The `acceptConnection` method is in a silly place in that the
`ServerSocketGroup` represents a collection of resources that are
capible of listening for socket connections. It is the wrong place to be
defining the logic for accepting or rejecting connections. This is
better done via changing the definition of `BufferPipelineBuilder` from
`SocketConnection => LeafBuilder[ByteBuffer]`
to
`SocketConnection => Option[LeafBuilder[ByteBuffer]]`
It also makes sense to rename `BufferPipelineBuilder` to
`SocketPipelineBuilder` since it must be dealing in `SocketConnection`s.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/ServerChannelGroup.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/ServerChannelGroup.scala
@@ -12,20 +12,11 @@ import scala.util.Try
   * @note Implementations may have resources associated with
   *       them before binding any sockets and should be closed.
   */
-abstract class ServerChannelGroup {
-  protected val logger = getLogger
+trait ServerChannelGroup {
 
   /** Create a [[ServerChannel]] that will serve the service on the specified socket */
-  def bind(address: InetSocketAddress, service: BufferPipelineBuilder): Try[ServerChannel]
+  def bind(address: InetSocketAddress, service: SocketPipelineBuilder): Try[ServerChannel]
 
   /** Closes the group along with all current connections. */
   def closeGroup(): Unit
-
-  /** Decide whether to accept the incoming connection or not
-    *
-    * Should also be used for any requisite logging purposes. */
-  protected def acceptConnection(address: InetSocketAddress): Boolean = {
-    logger.info(s"Connection from $address accepted at ${new Date}.")
-    true
-  }
 }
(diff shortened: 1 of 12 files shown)


bryce-anderson:
I had an epiphany on the way home from work: this operation should probably be asynchronous. That will allow the pipeline to asynchronously check if it should accept a socket to prevent DOS attacks (though in reality this should be done not-in-the-app) or collect resources it may need. It will almost certainly be a waste of a `Future` allocation in the majority of cases but its also only once per socket connection, so it should be a drop in the bucket.

bryce-anderson:
Alright, I changed this to `Future`, which I like much better.

rossabaker:
What is the proper way to reject?  A failed future?  Containing what?

bryce-anderson:
Good question. A failed future of whatever you wanted to say was good enough for me, but perhaps we should provide an explicit rejection `Exception`? I suspect the reasons for rejecting would be varied (and extraordinarily rare), but a reasonable base seems appropriate.

bryce-anderson:
Unfortunately we can't constrain what that rejection looks like. We could make it a `Future[Option[TailBuilder]]`, though things getting deep.

rossabaker:
Or `Future[Either[RejectionReason, TaillBuilder]]`, but yeah, getting uglier.  Ain't no `EitherT` down here.

I just wanted to understand the idiom, because it mostly looks like it adds a bunch of `Future.successful`.

bryce-anderson:
@rossabaker, based on your comments, I'm still a little dissatisfied with this as-is. 

I'm thinking about making a base exception, `ConnectionRejectedException` which represents the rejected case which will result in an info log and other exceptions will result in an error log. I'm also not against the `Future[Either[RejectionReason, TaillBuilder]]` idea if you think it's the right way to go, though if I had to flip a coin I'd personally hope it lands on the special exception.

Another thing to consider is to give control back to the user: instead of passing in a
```scala
SocketConnection => Future[LeafBuilder[ByteBuffer]]
```
They could pass in a
```scala
HeadStage[ByteBuffer] => Unit
```
and do whatever they want with it whenever they want. This would require refining the type of `HeadStage[ByteBuffer]` into something like 
```scala
SocketHead extends HeadStage[ByteBuffer]  {
  val connection: SocketConnection
}
```
and care would need to be taken to ensure that we don't try to send inbound events (something I'm thinking about with https://github.com/http4s/blaze/pull/190)
but honestly, that feels like the right API.

bryce-anderson:
I'm going to merge this as-is and we can continue to iterate on it as the need arises.

PR #192, opened 2018-03-19
Tails are in charge of closing resources

bryce-anderson:
Right now, when a channel fails to read or write it fails the channel,
closing everything etc. This makes it impossible to handle half-closed
scenarios. Instead of the Head being in charge of releasing resources,
we move that responsibility to the tails. This allows them to interact
with a half closed socket.

Note: this doesn't yet allow half-closing a socket, though it is a
necessary step.


rossabaker:
Reviews may be a bit tough to come by.  There are 7 committers in Boston right now.  I'll try to give this its due tomorrow, but life does not return to normal until Wednesday.

bryce-anderson:
@rossabaker no rush at all.

PR #195, opened 2018-03-24
Make HTTP message toString safe by default

bryce-anderson:
The HttpRequest and HttpResponsePrelude are case classes which will
expose the headers in the default `toString` method implementation.
Instead, we now implement a safe `toString` method but also provide an
`unsafeToString` method for cases where you really want them.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/util/Property.scala
+++ b/core/src/main/scala/org/http4s/blaze/util/Property.scala
@@ -1,0 +1,25 @@
+package org.http4s.blaze.util
+
+/** Helper for defining behavior specific to properties */
+private[blaze] abstract class Property(default: Boolean) {
+  // This may not be the most robust way to fix up the names of objects.
+  final val name = getClass.getName.replace("$", "")
+
+  /** Get the value of the property if it is set */
+  final def get: Option[Boolean] = System.getProperty(name) match {
+    case null => None
+    case value => Some(value == "true")
+  }
+
+  /** Get the value of this property, defaulting to `false` if it is not set. */
+  final def apply(): Boolean = System.getProperty(name) match {
+    case null => default
+    case value => value == "true"
+  }
+
+  /** Set the value of this property */
+  final def set(value: Boolean): Unit = {
+    System.setProperty(name, value.toString)
+    ()
+  }
+}
--- a/http/src/main/scala/org/http4s/blaze/http/http1/client/Http1ClientStage.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http1/client/Http1ClientStage.scala
@@ -15,7 +15,6 @@ import scala.concurrent.{Future, Promise}
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
-// TODO: we're totally logging all sorts of headers in here, which is bad for security
 private[http] class Http1ClientStage(config: HttpClientConfig)
     extends TailStage[ByteBuffer]
     with Http1ClientSession {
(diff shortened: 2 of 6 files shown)


aryairani:
Does it mean unsafe as in exposing some information to user potentially, or unsafe as in could crash?

bryce-anderson:
@aryairani, it's unsafe in that it potentially exposes sensitive data.

@rossabaker, I like the idea of a system property that defines this behavior. Do we have a similar property defined for for http4s proper? This kind of segues into the topic of having some sort of flags library that could be shared between blaze and http4s if we find enough of these situations arising.

rossabaker:
There are no such properties yet. There is a server logging trait being added to http4s-blaze that takes a Request and Response and gives most of the flexibility we need.  (It lacks timing and byte counts for various transfer encodings.)

The existing logging middleware in http4s-server takes a predicate to filter the headers. Maybe something like this could be passed to blaze, with a safe default.

Raw wire dumps sound like a pipeline atage outside blaze-http, maybe?

aryairani:
I’d like to nominate “debugToString” as an alternative to “unsafeToString”;  that is all.

bryce-anderson:
@aryairani, if we keep the two separate methods I'll rename it to `debugToString`.

@rossabaker, there is a little extra value in the raw dump where it's at in that it shows that the client is trying to interpret as the response. This could come into play if parsing the body causes problem and we find the byte stream in an invalid state. That said, it's pretty old and probably of little value at this stage, so maybe it's fine to just remove it.

bryce-anderson:
I took the flag approach, and renamed the verbose toString as `sensitiveToString` and I really like where it landed. More feedback welcome; maybe I'm the only one that likes the result. 😄

core/src/main/scala/org/http4s/blaze/util/Property.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  /** Get the value of this property, defaulting to `false` if it is not set. */
+  final def apply(): Boolean = System.getProperty(name) match {
+    case null => default
+    case value => value == "true"
+  }
+
+  /** Set the value of this property */
+  final def set(value: Boolean): Unit = {
+    System.setProperty(name, if (value) "true" else "false")
```
rossabaker: I think that if expression is just `value.toString`.

core/src/main/scala/org/http4s/blaze/util/Property.scala
```diff
@@ -0,0 +1,25 @@
+package org.http4s.blaze.util
+
+/** Helper for defining behavior specific to properties */
+private[blaze] abstract class Property(default: Boolean) {
+  // This may not be the most robust way to fix up the names of objects.
+  final val name = getClass.getName.replace("$", "")
+
+  /** Get the value of the property if it is set */
+  final def get(): Option[Boolean] = System.getProperty(name) match {
```
rossabaker: Is pure, doesn't need parens.

http/src/main/scala/org/http4s/blaze/http/http1/client/Http1ClientStage.scala
```diff
@@ -306,7 +311,9 @@ private[http] class Http1ClientStage(config: HttpClientConfig)
       new ClientBodyReader(buffer)
     }
 
-  /** The returned Future is satisfied once the entire request has been written
+  /** Begin writing a response to the peer.
+    *
+    * The returned Future is satisfied once the entire request has been written
```
rossabaker: s/satisfied/completed/ ?

PR #196, opened 2018-03-24
NIO1HeadStage can now potentially read and write without waiting on NIO notification

bryce-anderson:
Currently, reading and writing requires notification via NIO interests.
This can be really wasteful if there is already data in the inbound buffer
or space in the outbound buffer.

Now we try to read and write data initially, and only if we can't do we
fall back to interest operations.


bryce-anderson:
Anybody have any feedback on this? The only thing I'm hesitant about is the `LoopRunnable`: maybe it's just more noise than it's worth, but the alternative is to provide some other type of cached scratch buffer or thread the loops scratch buffer through a different way.

core/src/main/scala/org/http4s/blaze/channel/nio1/NIO1HeadStage.scala
```diff
@@ -214,22 +232,37 @@ private[nio1] final class NIO1HeadStage(
...
-          if (!BufferTools.checkEmpty(writes)) { // Non-empty buffer
-            writePromise = p
-            writeData = writes
-            setOp(SelectionKey.OP_WRITE)
-            p.future
-            ()
+          if (!BufferTools.checkEmpty(writes)) {
+            // Non-empty buffers. First we check to see if we can immediately
+            // write all the data to save a trip through the NIO event system.
+            performWrite(ch, scratch, writes) match {
```
hvesalai: I like this. All small HTTP responses can be written directly without NIO round trip.

bryce-anderson:
Cool, thanks for looking @ChristopherDavenport and @hvesalai. I'll merge this in the evening barring any last minute change requests.

PR #199, opened 2018-04-01
Remove the `Flush` command

bryce-anderson:
It isn't used anywhere other than `BufferingStage`, which is also
unused. Since we're trying to slim down and potentially remove the
`InboundCommand`s, this is a good place to start.


bryce-anderson:
I don't believe this is used in http4s proper.

Issue #198, opened 2018-04-01
Most outbound commands shouldn't be commands at all but members of the pipeline stages


bryce-anderson:
Things like `Disconnect` being represented as an `Event` diminish their importance, so we should explore making things like `def close(cause: Option[Throwable])` part of the `Stage` API.

Part of this may be to remove outbound commands altogether. The following known `OutboundCommand`s exist:
- `Connect`
- `Disconnect`
- `Flush`
- `Error`

`Flush` is essentially unused, error handling should be going the other way, and I think we can do a better with `Connect` via passing a pipeline builder to a connect call. This may not shake out to be true, but I think it's worth exploring.

bryce-anderson:
I think this could influence #197 since the `connect` mechanism is pretty much just a client concern.

bryce-anderson:
Closed via c87ff657dc49fd68267ec70faf0389260f988104.

PR #200, opened 2018-04-02
Remove the `Connect` command

bryce-anderson:
Related to #198

Only the `ClientChannelFactory` is making client channels, and they
start life connected, so essentially the `Connect` command is dead code.
Seems like a good time to remove it.


http/src/main/scala/org/http4s/blaze/http/http2/StreamStateImpl.scala
```diff
@@ -199,7 +199,7 @@ private abstract class StreamStateImpl(session: SessionCore) extends StreamState
   final override def outboundCommand(cmd: OutboundCommand): Unit =
     session.serialExecutor.execute(new Runnable {
       def run(): Unit = cmd match {
-        case Command.Flush | Command.Connect =>
+        case Command.Flush =>
```
rossabaker: This is going to conflict with your other PR, but I guess it's an obvious merge.

PR #202, opened 2018-04-28
Remove OutboundCommand

bryce-anderson:
Few of the defined types are useful other than `Disconnect`, which is
extremely important and could well be on the Stage itself.

We replace `Disconnect` and `Error` with a
`closePipeline(Option[Throwable)` method which subsumes both behaviors.


bryce-anderson:
Essentially, and `Error(Throwable)` is `Some(Throwable)`.

bryce-anderson:
It looks like there is more work to be done: my incremental compiler lied to me about core building. 😦

rossabaker:
Does `EOF` still have to be a `Throwable`?  Some of those cases are getting jostled by this.  Maybe that's still needed, or maybe that's out of scope, but I'm reminded that we're using an error for signaling.

bryce-anderson:
I wanted to remove `EOF`, but when I felt it was to large to do as part of this commit. Sounds like we're on the same page there.

bryce-anderson:
Maybe 'remove' was a bit much. It is pretty nice that our failure mode is distinguished from the 'good' half of `Future[T]` without needing to box it ,eg `def read(): Future[Option[T]]`. I do want to see what we can do though, because it is a bummer.

core/src/main/scala/org/http4s/blaze/channel/nio1/Selectable.scala
```diff
@@ -22,11 +22,5 @@ private trait Selectable {
...
     * This close method represents closure under normal conditions.
     */
-  def close(): Unit
-
-  /** Close this `Selectable` and release associated resources.
-    *
-    * This close method represents closure under exceptional conditions.
-    */
-  def closeWithError(cause: Throwable): Unit
+  def close(cause: Option[Throwable]): Unit
```
rossabaker: Is the scaladoc about "normal conditions" still true when there is some `cause`?

core/src/main/scala/org/http4s/blaze/channel/nio2/ByteBufferHead.scala
```diff
@@ -18,15 +18,15 @@ private[nio2] final class ByteBufferHead(channel: AsynchronousSocketChannel, buf
...
+  private[this] var closeReason: Option[Throwable] = None
   private[this] val scratchBuffer = ByteBuffer.allocateDirect(bufferSize)
 
   override def writeRequest(data: ByteBuffer): Future[Unit] =
     writeRequest(data :: Nil)
 
   override def writeRequest(data: Seq[ByteBuffer]): Future[Unit] = {
     val reason = closeReason
-    if (reason != null) Future.failed(reason)
+    if (reason.isDefined) Future.failed(reason.get)
```
rossabaker: Is this worth turning into a `match` to get rid of the `.get`?

core/src/main/scala/org/http4s/blaze/channel/nio2/ByteBufferHead.scala
```diff
@@ -106,23 +106,25 @@ private[nio2] final class ByteBufferHead(channel: AsynchronousSocketChannel, buf
...
-      reason == null
+      reason.isEmpty
     }
 
-    t match {
-      case EOF => logger.debug(s"closeWithError(EOF)")
-      case t => logger.error(t)("NIO2 channel closed with an unexpected error")
+    cause match {
+      case Some(t) => logger.error(t)("NIO2 channel closed with an unexpected error")
+      case None => logger.debug(s"closeWithError(EOF)")
```
rossabaker: This scaladoc seems iffy now: it's not the name of the method nor the argument anymore.

core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
```diff
@@ -255,12 +244,18 @@ sealed trait Head[O] extends Stage {
...
+        case t: Throwable =>
+          logger.error(t)("Exception caught when attempting inbound command")
+          closePipeline(Some(t))
+      }
     } else {
-      val e = new Exception("Cannot send inbound command on disconnected stage")
-      logger.error(e)("")
+      val msg = "Cannot send inbound command on disconnected stage"
+      val e = new Exception(msg)
+      logger.error(e)(msg)
```
rossabaker: Not new, but why are we logging and throwing the same exception here?  I guess it's the age old discussion of the ideal being exactly once logging and the counterargument of at-least once logging.

bryce-anderson: Old code since before we had that discussion. I think we should remove the log and just throw the exception. I'll change it.

bryce-anderson:
I think I've addressed most of the comments. Barring objections, I'll probably merge this before the end of the weekend.

PR #203, opened 2018-04-29
Add ability to close a BodyWriter with an exception

bryce-anderson:
Right now, there is no way for a user of a BodyWriter to signal to the
consuming side that it is being closed due to some problem, thus the
other end may think that the stream closed successfully, while in
reality it may be incomplete.

To fix this, we add a `cause: Option[Throwable]` to the `close` method
to allow the data producer to signal that it is closing due to a
problem.

### DIFF
--- a/http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
+++ b/http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
@@ -206,12 +206,14 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
     final override type Finished = RouteResult
 
     private[this] var closed = false
-    protected val lock: Object = this
+
+    final protected def lock: Object = this
 
     protected def doWrite(buffer: ByteBuffer): Future[Unit]
 
     protected def doFlush(): Future[Unit]
 
+    // Must only be called from while synchronized on `lock`
     protected def doClose(): Future[RouteResult]
 
     final override def write(buffer: ByteBuffer): Future[Unit] =
@@ -225,16 +227,25 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
       else doFlush()
     }
 
-    final override def close(): Future[RouteResult] = lock.synchronized {
+    final override def close(cause: Option[Throwable]): Future[RouteResult] = lock.synchronized {
       if (closed) InternalWriter.ClosedChannelException
       else {
         closed = true
-        doClose()
+        cause match {
+          case Some(ex) =>
+            // Since we're aborting, we need to just hang up for HTTP/1.x since
+            // we can't set a RST or signal in any other way that the response
+            // didn't terminate normally.
+            logger.debug(ex)("Closed due to exception")
+            FutureClose
+          case None =>
+            doClose()
+        }
       }
     }
   }
 
-  private class ClosingWriter(var sb: StringBuilder) extends InternalWriter {
+  private final class ClosingWriter(var sb: StringBuilder) extends InternalWriter {
 
     override def doWrite(buffer: ByteBuffer): Future[Unit] =
       if (sb == null) pipeline.channelWrite(buffer)
@@ -251,11 +262,11 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
 
     // This writer will always request that the connection be closed
     override def doClose(): Future[RouteResult] =
-      if (sb == null) Future.successful(Close)
+      if (sb == null) FutureClose
       else doFlush().map(_ => Close)(Execution.directec)
   }
 
-  private class FixedLengthBodyWriter(forceClose: Boolean, sb: StringBuilder, len: Long)
+  private final class FixedLengthBodyWriter(forceClose: Boolean, sb: StringBuilder, len: Long)
       extends InternalWriter {
 
     private var cache = new ArrayBuffer[ByteBuffer](4)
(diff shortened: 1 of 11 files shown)


bryce-anderson:
This continues the pattern of adding a `Option[Throwable]` to `close` methods. This kind of touches on what @rossabaker was commenting on in PR https://github.com/http4s/blaze/pull/202, which is that we have `EOF` which will always the same as `None`.

An alternative idea could be that we do `close() -> close(Throwable)` and have an overload `def close() = close(EOF)`. If we keep `EOF` around, this is probably a bit simpler in the long run, which is kind of what I'm leaning toward, but I'd like to have that debate before reworking this or PR https://github.com/http4s/blaze/pull/202.

http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
```diff
@@ -206,12 +206,14 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
...
     private[this] var closed = false
-    protected val lock: Object = this
+
+    final protected def lock: Object = this
 
     protected def doWrite(buffer: ByteBuffer): Future[Unit]
 
     protected def doFlush(): Future[Unit]
 
+    // Must be called from while holding the lock
```
aeons: Not 100% clear on what you mean here.

bryce-anderson: I tried to make it clearer, let me know if it's still a problem.

bryce-anderson:
It's not clear to me what the compounding effect of overloading `close()` is, especially in a world where `EOF` does still exist, since `None` is essentially a synonym for `EOF`.

I'm hesitant to remove `EOF` because in many situations, `EOF` is not terribly different than any other exception other than we may log the others more aggressively. For example, consider a server parser. This is because at certain points in execution, `EOF` _is_ actually exceptional (such as a peer hanging up mid message) and adding an extra wrapper `Either` bloats the logic and the heap. However, in other cases it's not.

http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
```diff
@@ -206,12 +206,14 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
     final override type Finished = RouteResult
 
     private[this] var closed = false
-    protected val lock: Object = this
+
+    final protected def lock: Object = this
```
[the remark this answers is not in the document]
bryce-anderson: There is no meaningful runtime difference, so just a personal quirk, I guess. 😄 The `lock` is shared between the parent and subclasses, and in those cases I personally find it distasteful to use `this` as an implicitly shared lock as opposed to an explicitly declared `lock`.

I'd actually like your opinion on the above. It's just something I've started doing, but if it seems ridiculous or counterintuitive, I may just stop doing it.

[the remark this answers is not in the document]
rossabaker: It's less clear because it's quirky but more clear because it's better named. I don't have a strong feeling about the net value.  I'm fine with it.

http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
```diff
@@ -225,16 +227,25 @@ private final class Http1ServerCodec(maxNonBodyBytes: Int, pipeline: TailStage[B
       else doFlush()
     }
 
-    final override def close(): Future[RouteResult] = lock.synchronized {
+    final override def close(cause: Option[Throwable]): Future[RouteResult] = lock.synchronized {
```
jmcardon: Can `close` possibly fail with something completely unexpected, or does it fail with a (relatively) enumerable set of exceptions?

Asking because if possible, I think `close(cause: Option[Exception])` forces me to squint a little less at this. `Throwable` encompasses way too much in my eyes, a lot of which is "your program is underfined now".

http/src/main/scala/org/http4s/blaze/http/http2/server/AbstractBodyWriter.scala
```diff
@@ -25,6 +23,8 @@ private abstract class AbstractBodyWriter(private var hs: Headers) extends BodyW
 
   protected def flushMessage(msg: Seq[StreamFrame]): Future[Unit]
 
+  protected def fail(cause: Throwable): Unit
```
jmcardon: Here the same as above: can we restrict this possibly so we don't go into throwable territory?

http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
bryce-anderson: It looks like we can refine it to `Exception`, and that seems reasonable. I think the only place that will cause issues if when someone attempts to close due to the failure half of a `Future`, but I suppose we can cross that bridge later.

rossabaker: Is this argument an offshoot of "all Throwables are fatal and all Exceptions are not?"  That's essentially the original JVM intent, but we have two competing definitions that contradict that in Scala (stdlib's `NonFatal` and fs2/cats-effect's `NonFatal`).  It would be nice if we had a consistent interpretation of fatality up and down the stack.

bryce-anderson: That's a reasonable point. I'm having a change of heart, and think we should probably move it back to `Throwable`.

rossabaker: It's unfortunate that `NonFatal` is an extractor rather than a type, or that a few accepted non-fatal things extend `Error`.  What @jmcardon argues for is in line with the original intent of the JVM, but we lost our way on that before there was even a Scala.

bryce-anderson: Alright, reverted back to `Throwable`. It's disappointing, but it is what it is.

http/src/main/scala/org/http4s/blaze/http/http1/server/Http1ServerCodec.scala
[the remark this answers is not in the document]
ChristopherDavenport: I actually like this. For once I understand whats going on with this notation even in imperative land.

PR #205, opened 2018-05-07
Change SSLStage thread safety strategy to using SerialExecutor

bryce-anderson:
The SSLStage currently uses synchronization to ensure thread safety.
This makes it hard to maintain and reason about, especially considering
it used Promises frequently. We can change it to use the
`SerialExecutionContext` which ensures single-threaded behavior without
the need for locks.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
+++ b/core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
@@ -4,17 +4,14 @@ import java.nio.ByteBuffer
 import javax.net.ssl.SSLEngineResult.Status
 import javax.net.ssl.SSLEngineResult.HandshakeStatus
 import javax.net.ssl.{SSLEngine, SSLException}
-
 import scala.annotation.tailrec
 import scala.collection.mutable.{ArrayBuffer, Buffer, ListBuffer}
 import scala.concurrent.{Future, Promise}
 import scala.util.{Failure, Success}
-
 import org.http4s.blaze.pipeline.MidStage
 import org.http4s.blaze.pipeline.Command.EOF
 import org.http4s.blaze.util
-import org.http4s.blaze.util.Execution._
-import org.http4s.blaze.util.{BufferTools, ThreadLocalScratchBuffer}
+import org.http4s.blaze.util.{BufferTools, Execution, SerialExecutionContext, ThreadLocalScratchBuffer}
 import org.http4s.blaze.util.BufferTools._
 
 private object SSLStage {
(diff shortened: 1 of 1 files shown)


bryce-anderson:
Unfortunately the diff is significantly bigger than it what was changes since some indentation changed with the removal of synchronization blocks. That, and the `syncWrite` method was effectively removed and we now just use `doWrite`.

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -31,202 +38,203 @@ final class SSLStage(engine: SSLEngine, maxWrite: Int = 1024 * 1024)
...
+    val p = Promise[Unit]
+    serialExec.execute(new Runnable { def run(): Unit = doWrite(data, p) })
+    p.future
+  }
+
   // All processes that start a handshake will add elements to the queue
-  private def inHandshake() = handshakeQueue.nonEmpty
+  private[this] def inHandshake: Boolean = handshakeQueue.nonEmpty
 
   // MUST be called inside synchronized blocks
```
jmcardon: This is no longer true

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -31,202 +38,203 @@ final class SSLStage(engine: SSLEngine, maxWrite: Int = 1024 * 1024)
...
-          logger.warn(t)("SSLException in SSL handshake")
-          handshakeFailure(t)
-
-        case t: Throwable =>
-          logger.error(t)("Error in SSL handshake")
-          handshakeFailure(t)
+  // Perform the SSL Handshake and then continue any pending operations.
+  // There should be at least one pending operation as this is only started
+  // after such an operation is stashed.
+  private[this] def sslHandshake(data: ByteBuffer, r: HandshakeStatus): Unit = {
```
jmcardon: What's the reason for `private` => `private[this]`?

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -249,98 +257,50 @@ final class SSLStage(engine: SSLEngine, maxWrite: Int = 1024 * 1024)
...
-            if (out.nonEmpty) { // need to write
-              channelWrite(out).onComplete {
-                case Success(_) => continueWrite(data, p)
-                case Failure(t) => p.tryFailure(t)
-              }(trampoline)
-            } else sslHandshake(readData, r)
-
-          case SSLFailure(t) => p.tryFailure(t); ()
-        }
+        case SSLFailure(t) => p.tryFailure(t); ()
```
jmcardon: Why `tryFailure` here instead of `failed`?

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
bryce-anderson: It's a more appropriate notion of private. Vanill `private` allows the SSLStage instances touch each others methods while `private[this]` restricts access to only `this` instance.

I think in practice 99.9% of usages of `private` in classes/traits should really be `private[this]`.

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
bryce-anderson: Old mistake. I'll change it.

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
bryce-anderson: Indeed. Will fix, thanks!

bryce-anderson:
Thanks for the review @jmcardon!

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
jmcardon: ah interesting. I actually didn't know that `private[this]` restricted it to the instance. I've never run into it since I've never tried calling another `private` from another one of the same object.

Thanks!

rossabaker: `private[this]` also is direc access to a field instead of through a method, which can save a few picoseconds here and there.

PR #211, opened 2018-07-09
add 'oraclejdk10' to Travis matrix

sullis:


bryce-anderson:
Looks like there was a change to the API of `ByteBuffer` that isn't playing well with scala.

rossabaker:
Oops, hadn't noticed this one.  Fixed by #213 and #214.

PR #220, opened 2018-09-07
Release v0.14.0-M4

### LINKED ISSUE #218: Stages which inherit from `Stage` spam the logs with unhandled `Disconnected` events.
When running any basic server with blaze on http4s 0.19.x the logs are spammed with 
```
[blaze-selector-0-1] WARN  o.h.b.p.s.QuietTimeoutStage - org.http4s.blaze.pipeline.stages.QuietTimeoutStage Stage: 30 seconds received unhandled inbound command: Disconnected
[blaze-selector-0-1] WARN  o.h.s.b.Http1ServerStage$$anon$1 - Http4sServerStage received unhandled inbound command: Disconnected
```
This appears to be because of this block of code in `core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala`

```
  def inboundCommand(cmd: InboundCommand): Unit = cmd match {
    case Connected => stageStartup()
    case _ => logger.warn(s"$name received unhandled inbound command: $cmd")
  }
```

Neither `QuietTimeoutStage` nor `Http1ServerStage` override this to handle `Disconnected`, and the handling for `Disconnected` was removed in this commit: https://github.com/http4s/blaze/commit/b4ef5f59dfeb8bf47dba2008aacf0cb98368ead0#diff-8b9a55120b4e25f24f77063998d50fe6L66

rossabaker:
Proposing this intermediate release to unblock http4s milestone.

Fixes #218 by quieting `Disconnected` events.  These are not handled by `Http1ServerStage`, `QuietTimeoutStage`, `SSLStage`, which causes at least 2 warnings on every request, 3 if SSL.  HTTP/2 and websockets are probably also affected.  Maybe we ignore these at our own peril.

Catching up to the demise of `OutboundCommand` is left for https://github.com/http4s/http4s/issues/2058, but we need a milestone now.

Backported #216 in order to avoid dependency warnings on log4s.

This should be merged back to master after release.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
+++ b/core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
@@ -63,6 +63,7 @@ sealed trait Stage {
     */
   def inboundCommand(cmd: InboundCommand): Unit = cmd match {
     case Connected => stageStartup()
+    case Disconnected => logger.debug("Received Disconnected command")
     case _ => logger.warn(s"$name received unhandled inbound command: $cmd")
   }
 }
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import BlazePlugin._
 organization in ThisBuild := "org.http4s"
 
 lazy val commonSettings = Seq(
-  version := "0.14.0-M3",
+  version := "0.14.0-M4",
   description := "NIO Framework for Scala",
   crossScalaVersions := Seq("2.11.11", scalaVersion.value),
   scalacOptions in (Compile, doc) ++= Seq("-no-link-warnings") // Suppresses problems with Scaladoc @throws links
--- a/project/BlazePlugin.scala
+++ b/project/BlazePlugin.scala
@@ -93,18 +93,18 @@ object BlazePlugin extends AutoPlugin {
 
   lazy val logbackClassic      = "ch.qos.logback"             %  "logback-classic"     % "1.2.3"
   lazy val twitterHPACK        = "com.twitter"                %  "hpack"               % "1.0.2"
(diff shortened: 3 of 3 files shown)


core/src/main/scala/org/http4s/blaze/pipeline/Stages.scala
```diff
@@ -63,6 +63,7 @@ sealed trait Stage {
     */
   def inboundCommand(cmd: InboundCommand): Unit = cmd match {
     case Connected => stageStartup()
+    case Disconnected => logger.debug("Received Disconnected command")
```
rossabaker: But timeout stage is going to send a disconnect outbound, and if we ignore the disconnected inbound, the tail won't actually shut down.  This isn't going to work, is it?

rossabaker: * The `NIO1HeadStage` gets the `Disconnect` and shuts the connection down in a timely fashion.
* The `Http1ServerStage` gets the `Disconnected` notice in a timely fashion, but does nothing.  
  * This might be a good place to cancel any running effect, which only the stage knows.
  * Should the tail shut down on `Disconnected` if it wasn't shut down already?  But only the stage knows that, too.

rossabaker: `stageShutdown` is idempotent.  It (probably) doesn't send outbound commands.  So why isn't `stageShutdown()` still the proper default response to `Disconnected`?

rossabaker:
Oops.  I created this branch as a target for a branch on my fork and just pushed to origin, which closed the discussion.

Please keep discussing.

rossabaker:
This doesn't show my latest.  Starting a new PR. :frowning:

PR #233, opened 2018-10-02
Upgrade Scala and JDK versions

rossabaker:


rossabaker:
This build failure is repeatable. Unsure if it's due to openjdk or due to 11, but I'm not touching oraclejdk11 with a ten-foot pole.

I also can't install openjdk11 on NixOS without some work, so help wanted.

rossabaker:
Still a problem.  No blaze tests fail on the openjdk build of the http4s/http4s repo. :thinking:

PR #234, opened 2018-10-02
Release v0.14.0-M5

rossabaker:
This will unblock http4s/http4s#2146.


rossabaker:
Releasing.  "It's just a milestone."

Issue #235, opened 2018-10-03
http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2017-7656


joan38:
The following dependency:
```
lazy val alpn_api            = "org.eclipse.jetty.alpn"     % "alpn-api"             % "1.1.3.v20160715"
```
Is subject to this vulnerability:
http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2017-7656

Whatever this means, we can't use http4s in my company because of a requirement to have a OWASP score lower than 3.9.

Anyway to upgrade to a version newer than 2016?

Thanks

rossabaker:
This was also brought up on #231.  I'm still unclear whether it's the alpn-api or the runtime that has the vulnerability.  I could not find a newer version of the api.  The alpn-boot jar, which is only used in examples, is upgraded in #232.  Can you clarify whether the API itself is compromised?

joan38:
@rossabaker Thanks for the hint.
Have a look at https://github.com/http4s/blaze/pull/236

joan38:
This is a false positive vulnerability. Solution:
```
<?xml version="1.0" encoding="UTF-8"?>
<suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.1.xsd">
    <suppress>
        <notes><![CDATA[file name: alpn-api-1.1.3.v20160715.jar]]></notes>
        <gav regex="true">^org\.eclipse\.jetty\.alpn:alpn-api:.*$</gav>
        <cpe>cpe:/a:eclipse:jetty</cpe>
    </suppress>
</suppressions>
```

```
```

sbordet:
I'm the `org.eclipse.jetty.alpn` author.

@joan38 vulnerability CVE-2017-7656 has been addressed here: https://www.eclipse.org/lists/jetty-announce/msg00123.html.

One thing is the Jetty server itself, and another thing are jars of libraries that happen to have the same or a similar `groupId` as in `org.eclipse.jetty`.

CVE-2017-7656 is about a behavior in the Jetty HTTP parser, so `org.eclipse.jetty.alpn` is completely unaffected by it - I guess the vulnerability tools got confused.

joan38:
Thanks @sbordet!
I've reported this false positive to https://github.com/jeremylong/DependencyCheck/issues/1515

PR #236, opened 2018-10-03
Drop alpn vulnerable dependency

joan38:
Attempt to solve https://github.com/http4s/blaze/issues/235


joan38:
```
[warn] blaze-http >>> The following libraries are declared in libraryDependencies but are not needed for compilation:
[warn]  - "org.eclipse.jetty" % "jetty-alpn-openjdk8-client" % "9.4.12.v20180830"
[info] Compiling 8 Scala sources to /home/travis/build/http4s/blaze/examples/target/scala-2.12/classes ...
[info] Done compiling.
[info] blaze-examples >>> The project has no unused dependencies declared in libraryDependencies. Good job!
[error] (blaze-http / unusedCompileDependenciesTest) Failing the build because unused dependencies were found
```
This is not true. If the lib is remove it won't compile.
What can we do?

rossabaker:
jetty-alpn-openjdk8-client is not directly used, which is the reason for this error.  That's intended to be a runtime dependency at application level, not a library dependency.

The reason that removing jetty-alpn-openjdk8-client fails to compile is that it transtively pulls in alpn-api, which _is_ a compile time dependency.

I am still unclear whether there is a vulnerability in the API, as opposed to the implementation.  See comment on #235.

rossabaker:
What dependency checker is blacklisting this?  My question remains: is the vulnerability in the _API_ or in the _implementation_?  The CVEs refer to older versions of Jetty.  There is no newer version of this API.  As far as I can tell, the vulnerability is in older versions of the implementation.

joan38:
sbt-dependency-check, look at the code I added the plugin to try out and it seems that `Provided` is not making him happy.
I see what you mean I don't know why this is flagged like this.

joan38:
I found a project excluding it:
https://github.com/Opetushallitus/koski/blob/master/owasp-dependency-check-suppressions.xml

joan38:
Alright I think we can close this. The solution is to add this false positive exception in that XML file.

PR #242, opened 2018-10-16
Don't warn on the removal of a disconnected MidStage

rossabaker:
It's idempotent, and so is stage shutdown. No reason to scare people in the logs.


rossabaker:
I never use trace.  I'd be willing to bring it back as debug.

ChristopherDavenport:
If it has no value, we should remove it. If it might have value when looking at internals decreasing the log level seems appropriate. We can always bring it back later.

rossabaker:
I don't think it's interesting at all, but I may have a narrow use case in mind.  I'll add debug logging.

rossabaker:
Build failure is the disappearance of oraclejdk10, addressed in #233.

PR #243, opened 2018-10-16
Cache Future.successful(()) and Future.failure(EOF)

rossabaker:


rossabaker:
Build failure is the disappearance of oraclejdk10, addressed in #233.

PR #244, opened 2018-10-20
Release v0.14.0-M8

rossabaker:
Includes two build corrections:
- We specified the version in two places
- scalaVersion was not set properly if not in Travis


ChristopherDavenport:
Does this deploy process go through travis?

rossabaker:
No.  It has rig, but we never finished setting it up.

PR #246, opened 2018-10-21
Tweaks in the fight against http4s/http4s#2192

rossabaker:


core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -135,11 +135,15 @@ final class SSLStage(engine: SSLEngine, maxWrite: Int = 1024 * 1024)
...
     val start = System.nanoTime
-    val results = handshakeQueue.result(); handshakeQueue.clear();
-    results.foreach {
-      case DelayedRead(_, p) => p.failure(t)
-      case DelayedWrite(_, p) => p.failure(t)
-    }
+    val results = handshakeQueue.result()
+    handshakeQueue.clear()
+    if (results.isEmpty)
+      logger.error(t)(s"Handshake failure with no pending results")
```
rossabaker: I'm not sure that this should ever happen, but we shouldn't be silent about it if it does.

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -182,15 +186,15 @@ final class SSLStage(engine: SSLEngine, maxWrite: Int = 1024 * 1024)
           o.flip()
 
           if (r.bytesProduced < 1)
-            logger.warn(s"SSL Handshake WRAP produced 0 bytes.\n$r")
+            handshakeFailure(new SSLException(s"SSL Handshake WRAP produced 0 bytes: $r"))
```
rossabaker: This is a loop we're getting into in http4s/http4s#1492. Let's see what happens if we fail it aggressively.

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -182,15 +186,15 @@ final class SSLStage(engine: SSLEngine, maxWrite: Int = 1024 * 1024)
...
 
           channelWrite(copyBuffer(o)).onComplete {
             case Success(_) => sslHandshake(data, r.getHandshakeStatus)
             case Failure(t) => handshakeFailure(t)
           }(serialExec)
 
         // Finished with the handshake: continue what we were doing.
-        case _ =>
+        case HandshakeStatus.FINISHED =>
```
rossabaker: I think this is always true.  Let's find out.

PR #251, opened 2018-11-08
A handshake can have a status of NOT_HANDSHAKING after all

rossabaker:
Fixes issue http4s/http4s#2259, I think.


rossabaker:
This is the way it had been for a long time and I thought the case was too broad but it turns out I was wrong.

PR #280, opened 2019-04-17
Upgrade to Scala 2.13.0

mcanlas:


mcanlas:
Blocked by

```
[error] unresolved dependency: org.log4s#log4s_2.13.0-RC1;1.7.0: not found
[error] unresolved dependency: org.scoverage#scalac-scoverage-plugin_2.13.0-RC1;1.4.0-M5: not found
```

mcanlas:
Links

- https://github.com/Log4s/log4s/pull/32
- https://github.com/scoverage/scalac-scoverage-plugin/issues/250

rossabaker:
Thanks for getting this rolling!

I think we can drop scoverage.  We dropped it on http4s core.  Tests are important, but nobody looks at the reports.

rossabaker:
I decided to keep 2.13.0-M5 support, just in case we need to patch blaze again before http4s-0.20 is EOL.  It mildly complicated a few things.  We could just as well have bumped to blaze-0.15, but nothing significant changed other than adding a new crossbuild.

aeons:
Looks like it's still pulling in an incompatible version of scoverage through rig.

aeons:
There's no sbt-scoverage for 2.13.0-M5 it seems.

rossabaker:
Meh. I thought we had ripped rig out.  I'll see what I can do.

aeons:
I think we all agreed it would be a good idea to do so :D

ChristopherDavenport:
I ripped it out of http4s, not blaze. Could likely find the PR and recreate though fairly easily.

rossabaker:
Releasing as v0.14.5.

rossabaker:
Ugh, we lost `publishTo`.

PR #288, opened 2019-05-02
Add "The scpecified network name is no longer available\r\n" to the brokenPipeMessages.

RafalSumislawski:
By mistake I reported the issue in http4s, more details available there: https://github.com/http4s/http4s/issues/2496

Fixes: https://github.com/http4s/http4s/issues/2496


rossabaker:
We can do an 0.14.1 release of blaze, and then get that into 0.20.1 of http4s.

rossabaker:
Nuts. I meant to release this with http4s-0.20.1. I'll work on releasing this and you can manually upgrade, and we'll make sure http4s-0.20.2 includes it.

PR #289, opened 2019-05-04
sbt-sonatype 2.5

eed3si9n:
I was testing the latest sbt 1.3.0 milestone on Community Build and found that Blaze won't load up because sbt-sonatype 1.1 is not available in Maven Central.


rossabaker:
Thanks!

PR #295, opened 2019-06-10
Don't catch fatal exceptions

### LINKED ISSUE #294: Blaze catches fatal exceptions
I'm debugging an OOM exception happening on moderate/heavy load and seeing that http4s blaze server with nio2 enabled is catching fatal errors preventing JVM from crash.

Here's an example traceback I'm getting:

```
[scala-execution-context-global-37] INFO org.http4s.server.blaze.Http1ServerStage$$anon$1 - HTTP body not read to completion. Dropping connection.
[scala-execution-context-global-39] INFO org.http4s.server.blaze.Http1ServerStage$$anon$1 - HTTP body not read to completion. Dropping connection.
[blaze-selector-1] ERROR org.http4s.blaze.util.Execution - Trampoline EC Exception caught
    java.lang.OutOfMemoryError: Direct buffer memory
    at java.nio.Bits.reserveMemory(Bits.java:695)
    at java.nio.DirectByteBuffer.<init>(DirectByteBuffer.java:123)
    at java.nio.ByteBuffer.allocateDirect(ByteBuffer.java:311)
    at org.http4s.blaze.channel.nio2.ByteBufferHead.<init>(ByteBufferHead.scala:22)
    at org.http4s.blaze.channel.nio2.NIO2SocketServerGroup$NIO2ServerChannel$$anon$1.$anonfun$completed$1(NIO2SocketServerGroup.scala:118)
    at org.http4s.blaze.channel.nio2.NIO2SocketServerGroup$NIO2ServerChannel$$anon$1.$anonfun$completed$1$adapted(NIO2SocketServerGroup.scala:114)
    at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
    at org.http4s.blaze.util.Execution$ThreadLocalTrampoline.run(Execution.scala:108)
    at org.http4s.blaze.util.Execution$ThreadLocalTrampoline.execute(Execution.scala:94)
    at org.http4s.blaze.util.Execution$$anon$2.execute(Execution.scala:53)
    at scala.concurrent.impl.CallbackRunnable.executeWithValue(Promise.scala:72)
    at scala.concurrent.impl.Promise$KeptPromise$Kept.onComplete(Promise.scala:372)
    at scala.concurrent.impl.Promise$KeptPromise$Kept.onComplete$(Promise.scala:371)
    at scala.concurrent.impl.Promise$KeptPromise$Successful.onComplete(Promise.scala:379)
    at org.http4s.blaze.channel.nio2.NIO2SocketServerGroup$NIO2ServerChannel$$anon$1.completed(NIO2SocketServerGroup.scala:125)
    at org.http4s.blaze.channel.nio2.NIO2SocketServerGroup$NIO2ServerChannel$$anon$1.completed(NIO2SocketServerGroup.scala:110)
    at sun.nio.ch.Invoker.invokeUnchecked(Invoker.java:126)
    at sun.nio.ch.Invoker$2.run(Invoker.java:218)
    at sun.nio.ch.EPollPort$EventHandlerTask.run(EPollPort.java:287)
    at sun.nio.ch.AsynchronousChannelGroupImpl$1.run(AsynchronousChannelGroupImpl.java:112)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
```

As far as I can see, this is where exceptions are caught: https://github.com/http4s/blaze/blob/92ad9100edda7585d6d7e99e940d1de58ba6768b/core/src/main/scala/org/http4s/blaze/util/Execution.scala#L109

After this exception server runs into an invalid state, stops responding to HTTP requests, but still haning in the memory and keeps trying to run the runnables resuling in the same traceback.

rossabaker:
Fixes #294

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/util/Execution.scala
+++ b/core/src/main/scala/org/http4s/blaze/util/Execution.scala
@@ -8,6 +8,7 @@ import org.log4s.getLogger
 import scala.annotation.tailrec
 import scala.concurrent.duration.Duration
 import scala.util.Try
+import scala.util.control.NonFatal
 
 /** Special `ExecutionContext` instances.
   */
@@ -106,7 +107,7 @@ object Execution {
         running = false
       } else {
         try r.run()
-        catch { case e: Throwable => reportFailure(e) }
+        catch { case NonFatal(t) => reportFailure(t) }
         run()
       }
     }
--- a/core/src/main/scala/org/http4s/blaze/util/Execution.scala
+++ b/core/src/main/scala/org/http4s/blaze/util/Execution.scala
@@ -107,7 +107,14 @@ object Execution {
         running = false
       } else {
         try r.run()
-        catch { case NonFatal(t) => reportFailure(t) }
+        catch {
+          case NonFatal(t) => reportFailure(t)
+          case t: Throwable =>
+            // In case someone catches t against all advice of
+            // NonFatal, leave the trampoline in a good state.
+            next()
+            throw t
+        }
         run()
       }
     }
--- a/core/src/main/scala/org/http4s/blaze/util/Execution.scala
+++ b/core/src/main/scala/org/http4s/blaze/util/Execution.scala
@@ -109,11 +109,6 @@ object Execution {
         try r.run()
         catch {
           case NonFatal(t) => reportFailure(t)
-          case t: Throwable =>
-            // In case someone catches t against all advice of
-            // NonFatal, leave the trampoline in a good state.
-            next()
-            throw t
         }
         run()
       }
--- a/core/src/test/scala/org/http4s/blaze/util/ExecutionSpec.scala
+++ b/core/src/test/scala/org/http4s/blaze/util/ExecutionSpec.scala
@@ -1,16 +1,17 @@
 package org.http4s.blaze.util
 
 import org.specs2.mutable.Specification
-import Execution.{directec, trampoline}
 
 class ExecutionSpec extends Specification {
+  private def trampoline = Execution.trampoline
+
   def toRunnable(f: => Unit): Runnable = new Runnable {
     override def run(): Unit = f
   }
 
-  "trampoline" should {
-    def submit(f: => Unit): Unit = trampoline.execute(toRunnable(f))
+  def submit(f: => Unit): Unit = trampoline.execute(toRunnable(f))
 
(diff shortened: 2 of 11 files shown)


core/src/test/scala/org/http4s/blaze/util/ExecutionSpec.scala
```diff
@@ -89,5 +88,21 @@ class ExecutionSpec extends Specification {
 
       i must be equalTo(iterations)
     }
+
+    "not catch fatal exceptions" in {
```
rossabaker: This is the one that was reported.  Testing the others would be tedious, but welcome if someone is so ambitious.

core/src/main/scala/org/http4s/blaze/util/Execution.scala
```diff
@@ -106,7 +107,14 @@ object Execution {
         running = false
       } else {
         try r.run()
-        catch { case e: Throwable => reportFailure(e) }
+        catch {
+          case NonFatal(t) => reportFailure(t)
+          case t: Throwable =>
+            // In case someone catches t against all advice of
+            // NonFatal, leave the trampoline in a good state.
+            next()
```
rossabaker: I believe that this should not allocate memory or otherwise mask `t`, though I have not studied the implementation of `ArrayDeque`.  We could arguably do without this, but we would have to remove the new test, which otherwise leaves the ThreadLocal in a state that breaks subsequent tests in a flaky, enraging fashion.

Alternatively, we could clear the ThreadLocal in a catch block that targets fatals in `trampoline#execute`, though we'd then potentially lose other queued tasks and deadlock.

Or we could just not test that fatal is fatal.

rossabaker: Well, that didn't work.  Let's see what happens [without the fatal test](https://travis-ci.org/http4s/blaze/builds/544050450), but I'm going to bed.

rossabaker: I don't think there's anything sensible we can do here.  Any that are in the queue have already been submitted and expect to run, but we need to throw, so we can't run the rest.  Fatal errors are fatal to the trampoline.

PR #300, opened 2019-06-20
Switch to official scalafmt plugin

aeons:
Enable checking formatting in CI and...

actually run scalafmt on the source :)


rossabaker:
This is going to be hell on our history, but every old project goes through this once.

PR #305, opened 2019-07-08
SSL Engine cleanup

balatbn:
Cleanup SSL Engine during connection close. 

With JDK SSLContext, this is not required. With native ssl context(Ex: conscrypt), calling close() methods frees up native memory associated with each connection. Otherwise, system will leak memory for every connection.


balatbn:
This throws SSLException. Is it OK if I just handle SSLException here?

PR #308, opened 2019-07-21
Implement connecting timeout in ClientChannelFactory.

RafalSumislawski:
This PR is motivated by repeating issues around connecting timeouts in http4s:
- https://github.com/http4s/http4s/issues/2386
- https://github.com/http4s/http4s/issues/2338
- https://github.com/http4s/http4s/issues/2690

The current state of connecting timeouts in http4s is that a timeout should happen eventually, but **when** it will happen is up to the OS. For example I've measured 15 seconds on Windows 7 and 135 seconds on Ubuntu 16.04, and @kamilkloch  (https://github.com/http4s/http4s/issues/2690) reports that in his environment the timeout doesn't happen at all. 

For comparison other http client libraries let users configure the timeout: 
- [`akka-http`](https://github.com/akka/akka-http/blob/eb7311130021c6fa72dff05090c20568ee1319dd/akka-http-core/src/main/resources/reference.conf#L256),
- [`sttp`](https://github.com/softwaremill/sttp/blob/cfc757f7449e992aac39af883cbdc0c7e22004a0/docs/conf/timeouts.rst#L6)

Establishing connections is responsibility of `ClientChannelFactory`, so that's where I start the work. I plan to follow up with a PR to http4s, bringing the timeout configuration to the surface of `BlazeClientBuilder`.

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
@@ -27,6 +27,13 @@ final class ClientChannelFactory(
     scheduler: TickWheelExecutor = Execution.scheduler,
     connectingTimeout: Option[FiniteDuration] = None) {
 
+  // for binary compatibility with <=0.14.6
+  def this(
+      bufferSize: Int,
+      group: Option[AsynchronousChannelGroup],
+      channelOptions: ChannelOptions) =
+    this(bufferSize, group, channelOptions, Execution.scheduler, None)
+
   def connect(
       remoteAddress: SocketAddress,
       bufferSize: Int = bufferSize): Future[HeadStage[ByteBuffer]] = {
--- a/core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
@@ -3,10 +3,12 @@ package org.http4s.blaze.channel.nio2
 import org.http4s.blaze.pipeline.HeadStage
 import java.nio.ByteBuffer
 import java.nio.channels.{AsynchronousChannelGroup, AsynchronousSocketChannel, CompletionHandler}
-import java.net.SocketAddress
+import java.net.{SocketAddress, SocketTimeoutException}
 
 import org.http4s.blaze.channel.ChannelOptions
+import org.http4s.blaze.util.{Execution, TickWheelExecutor}
 
+import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{Future, Promise}
 import scala.util.control.NonFatal
 
@@ -21,7 +23,9 @@ import scala.util.control.NonFatal
 final class ClientChannelFactory(
     bufferSize: Int = DefaultBufferSize,
     group: Option[AsynchronousChannelGroup] = None,
-    channelOptions: ChannelOptions = ChannelOptions.DefaultOptions) {
+    channelOptions: ChannelOptions = ChannelOptions.DefaultOptions,
+    scheduler: TickWheelExecutor = Execution.scheduler,
+    connectingTimeout: Option[FiniteDuration] = None) {
 
   def connect(
       remoteAddress: SocketAddress,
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
```diff
@@ -21,7 +23,16 @@ import scala.util.control.NonFatal
 final class ClientChannelFactory(
     bufferSize: Int = DefaultBufferSize,
     group: Option[AsynchronousChannelGroup] = None,
-    channelOptions: ChannelOptions = ChannelOptions.DefaultOptions) {
+    channelOptions: ChannelOptions = ChannelOptions.DefaultOptions,
+    scheduler: TickWheelExecutor = Execution.scheduler,
+    connectingTimeout: Option[FiniteDuration] = None) {
+
+  // for binary compatibility with <=0.14.6
```
rossabaker: Nicely done.  (This incidentally is why http4s uses the builder pattern a lot in its backends.  It makes adding new options easier.)

core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
```diff
@@ -21,7 +23,16 @@ import scala.util.control.NonFatal
 final class ClientChannelFactory(
     bufferSize: Int = DefaultBufferSize,
     group: Option[AsynchronousChannelGroup] = None,
-    channelOptions: ChannelOptions = ChannelOptions.DefaultOptions) {
+    channelOptions: ChannelOptions = ChannelOptions.DefaultOptions,
+    scheduler: TickWheelExecutor = Execution.scheduler,
+    connectingTimeout: Option[FiniteDuration] = None) {
```
rossabaker: Why `Option[FiniteDuration]` instead of `Duration`, so no timeout is `Duration.Inf`?  (I wish we could eliminate negative durations in either encoding, but nobody tends to send them in practice.)

core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
```diff
@@ -30,18 +41,33 @@ final class ClientChannelFactory(
 
     try {
       val ch = AsynchronousSocketChannel.open(group.orNull)
+
+      val scheduledTimeout = connectingTimeout.map { t =>
+        val onTimeout = new Runnable {
+          override def run(): Unit = {
+            val finishedWithTimeout = p.tryFailure(new SocketTimeoutException())
```
rossabaker: Would it be useful to include the remote address and maybe the timeout in a message?

core/src/test/scala/org/http4s/blaze/channel/nio2/ClientChannelFactorySpec.scala
```diff
@@ -0,0 +1,32 @@
...
+      val factory =
+        new ClientChannelFactory(connectingTimeout = Some(1.millisecond), scheduler = scheduler)
+      val address = new InetSocketAddress("1.1.1.1", 1)
+
+      Await.result(factory.connect(address), 500.milliseconds) should throwA[SocketTimeoutException]
+    }
+  }
+}
+
+trait fastTickWheelExecutor extends After {
```
rossabaker: Maybe I'd move this into a util or test package so it's visible to other tests.

core/src/test/scala/org/http4s/blaze/channel/nio2/ClientChannelFactorySpec.scala
```diff
@@ -0,0 +1,32 @@
...
+import scala.concurrent.Await
+import scala.concurrent.duration.DurationDouble
+
+class ClientChannelFactorySpec extends Specification {
+
+  "ClientChannelFactory" should {
+    "time out" in new fastTickWheelExecutor {
+      val factory =
+        new ClientChannelFactory(connectingTimeout = Some(1.millisecond), scheduler = scheduler)
+      val address = new InetSocketAddress("1.1.1.1", 1)
```
rossabaker: That address is a CloudFlare service.  It doesn't listen on port 1, so it works, and probably always will, but I'm always reluctant to send traffic to servers that aren't ours. IPv6 [reserves `0100::/64` as a blackhole](https://tools.ietf.org/html/rfc6666).  I don't know of anything similar in IPv4, though there are a few [test nets](https://tools.ietf.org/html/rfc5737) that _probably_ behave similarly.

RafalSumislawski:
I agree with all your remarks. I've pushed a commit with the requested changes.

RafalSumislawski:
@rossabaker what do you think about exposing TickWheelExecutor.tick in public API? (for example by making it `val tick: Duration`). It would be useful for logging warning (in BlazeClientBuilder?) in case someone uses small `connectingTimeout` and large `tick`.

rossabaker:
Syncing to Maven Central as 0.14.7 now.

PR #312, opened 2019-07-31
Run scalafmt with the new version

aeons:
and remove the old, unused plugin.


aeons:
This got bigger than it looked in the beginning. I'll see if we can configure scalafmt to do something close to what we had before.

aeons:
I changed the config to make the diff smaller.

http/src/main/scala/org/http4s/blaze/http/http2/bits.scala
```diff
@@ -15,7 +15,7 @@ private[http2] object bits {
 
   object Masks {
     val INT31: Int = 0x7fffffff
-    val INT32: Long = 0xffffffffL
+    val INT32: Long = 0XFFFFFFFFL
```
rossabaker: It's curious it doesn't capitalize the `Int` literals, too.

core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
```diff
@@ -48,7 +48,9 @@ final class ClientChannelFactory(
             s"An attempt to establish connection with $remoteAddress timed out after $connectTimeout.")
           val finishedWithTimeout = p.tryFailure(exception)
           if (finishedWithTimeout) {
-            try { ch.close() } catch { case NonFatal(_) => /* we don't care */ }
+            try {
+              ch.close()
+            } catch { case NonFatal(_) => /* we don't care */ }
```
rossabaker: Do we know why it's more aggressively wrapping things now?  It's fine, but I shudder to think how large the diff is going to be on http4s core.

http/src/main/scala/org/http4s/blaze/http/http2/bits.scala
aeons: Yea, that's weird.

core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
aeons: I'll see if I can get the config to leave those alone.

aeons:
Haven't really had any luck with the oneline lambdas.

core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
aeons: https://github.com/scalameta/scalafmt/issues/1437

aeons: Yeah, that link was meant for the other comment.

Issue #339, opened 2019-10-03
Socket not closed if address is unresolvable (UnresolvedAddressException)


vanhoeij:
I think that I found an issue in the `org.http4s.blaze.channel.nio2.ClientChannelFactory`. It looks like the channel is not closed when an `UnresolvedAddressException` is thrown in the `connect()` function.

PR #340, opened 2019-10-03
Close channel if address is unresolvable or its type is not supported

vanhoeij:
Issue: https://github.com/http4s/blaze/issues/339

### DIFF
--- a/core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
+++ b/core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
@@ -56,22 +56,28 @@ final class ClientChannelFactory(
       }
       val scheduledTimeout = scheduler.schedule(onTimeout, connectTimeout)
 
-      ch.connect(
-        remoteAddress,
-        null: Null,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/org/http4s/blaze/channel/nio2/ClientChannelFactory.scala
```diff
@@ -56,22 +56,28 @@ final class ClientChannelFactory(
...
+              p.trySuccess(new ByteBufferHead(ch, bufferSize = bufferSize))
+              scheduledTimeout.cancel()
+            }
           }
-        }
-      )
+        )
+      } catch {
+        case ex: IllegalArgumentException =>
+          ch.close()
```
rossabaker: What if this throws?  I think we should catch and log here so we don't mask the original exception.

vanhoeij: Thanks for the review and good point about the exception handling! I have updated the code. I am not sure whether the logger is included correctly. It looks like it is done this way in other parts of the code.

PR #352, opened 2019-10-24
Add workaround for lightbend/mima#422

aeons:


rossabaker:
Still seems to be failing.

plokhotnyuk:
shouldn't it be defined in `project/plugins.sbt`?

aeons:
Yeah, that probably makes sense.

plokhotnyuk:
@aeons [here](https://github.com/plokhotnyuk/jsoniter-scala/blob/master/project/plugins.sbt#L2) is a more concise line that works for me as a W/A for that problem.

aeons:
Hah, yeah I was wondering why he would use the long-form of the resolver, but I thought that was part of the workaround.

aeons:
Ah, and it seems the entire issue is resolved now anyway :D

PR #354, opened 2019-11-14
Add trailing CRLF to chunks in ChunkedTransferEncoder

mvv:
As per the [RFC](https://tools.ietf.org/html/rfc7230#section-4.1). Found it when I was trying to connect blaze `HttpClient` to a http4s server.

Please release a new version with this fix.


rossabaker:
Released as 0.14.10

PR #407, opened 2020-05-10
sbt-http4s-org-0.1.0

rossabaker:
Test driving the new build plugin on a second repository.


rossabaker:
This is a real failure. It needs a MiMa exception, because blaze-0.14.0 didn't exist for Scala 2.13.

PR #409, opened 2020-05-11
Compiles clean under -Xlint:all

rossabaker:


rossabaker:
I left the fallthroughs since they seem to be by intentional: instead of looping back and reswitching, the code proceeds.  I am not sure whether we're saving nanoseconds here or whether the compiler optimizes that away anyway, but it's been working for a long time and I didn't want to touch it.

PR #418, opened 2020-06-11
add "connection reset" message thrown by java 13

edmothershaw:


edmothershaw:
I agree @rossabaker. However, I think the log may be useful for debugging in some situations, how about moving it to a debug line instead?

rossabaker:
So change line 25 to a debug, and eliminate line 23? Seems like a good idea to me.

PR #491, opened 2021-01-29
sbt-http4s-org-0.7

rossabaker:
A step toward Dotty support.

Inlined the unnecessary internal build plugins.  Removed unnecessary imports.  Added necessary headers.


rossabaker:
There's a big 2020->2021 pending if we don't merge this and it's getting annoying.

PR #492, opened 2021-01-29
Scala 3.0.0-M2 and 3.0.0-M3 support

rossabaker:
Builds on #491.

Some warnings from the type projection in `RouteAction` that I can't figure out how to eliminate, but the tests pass.  If the continuation were a SAM, I think we could refer to it as a path-dependent type.  We'd have to break compatibility to do it, which is an option for http4s-0.22.  http4s doesn't use it anyway, so this is a raw-blaze problem.

Some ominous errors from an existential type I can't find.  Again, tests pass.

```
[warn] An existential type that came from a Scala-2 classfile cannot be
[warn] mapped accurately to to a Scala-3 equivalent.
[warn] original type    : Seq[Class[? >: _$1 & _$1]] forSome _$1; _$1
[warn] reduces to       : Seq[Class[? >: _$1 & _$1]]
[warn] type used instead: Seq[Class[? >: Any]]
[warn] This choice can cause follow-on type errors or hide type errors.
[warn] Proceed at own risk.
```


rossabaker:
On that subject, we blocked #486 until we "got serious about Dotty."  Maybe this really is 0.15.

rossabaker:
The existential error is coming from StageToolsSpec, and the existential therein is not obvious to me.  I think it's Mockito.  We can live with this, but we may have to disable fatal warnings.

rossabaker:
Nope, it happens from all the Mockito tests.  I ain't touching those.

rossabaker:
So f0abb5e works on Scala 3 but not Scala 2. :tired_face:

rossabaker:
We're solving the generalized type projection by deleting the code that has the generalized type projection. :shrug:

PR #495, opened 2021-02-01
Prune blaze-http down to http4s' needs

rossabaker:
This discards an awful lot of interesting work, but leaves appropriate size dependency for http4s, which has been the dominant downstream user for years.  More importantly, it's a path to Scala 3 support, which `RouteAction` currently blocks.

We invite anyone still interested in blaze-without-http4s to revive from the 0.14.x line and discuss whether it makes sense to share maintenance or to fork.


rossabaker:
What's left is mostly HTTP/2 support.  It's unknown how many http4s-blaze-server users are on HTTP/2.  The number is probably small, but they've been given no warning of loss of support.

PR #499, opened 2021-02-06
Upgrade to 2.12.13, 2.13.4; fix CI crossbuilds

rossabaker:


http/src/main/scala/org/http4s/blaze/http/http2/PriorKnowledgeHandshaker.scala
```diff
@@ -65,6 +65,14 @@ abstract class PriorKnowledgeHandshaker[T](localSettings: ImmutableHttp2Settings
   // Attempt to read a SETTINGS frame and return it and any leftover data
   private[this] def readSettings(acc: ByteBuffer): Future[(MutableHttp2Settings, ByteBuffer)] = {
     logger.debug(s"receiving settings. Available data: $acc")
+
+    def insufficientData = {
```
RaasAhsan, line 69: Is there any logic changing here or just a slight refactor?

RaasAhsan (approved):

http/src/main/scala/org/http4s/blaze/http/http2/PriorKnowledgeHandshaker.scala
rossabaker: Any logic change is purely unintentional!

The old helper returned true on None, but Scala 2.13.4's improved checker didn't know that, so I added the case and extracted the duplicate rhs.

PR #517, opened 2021-04-02
Out with 3.0.0-M3, in with Scala 3.0.0-RC2

VlachJosef:
Helping with https://github.com/http4s/http4s/pull/4681


rossabaker (approved):
:tada:

PR #524, opened 2021-04-16
Bind ChannelSpec channels to localhost

rossabaker:
This should not make a difference, but similar did at a similar time in the main repo.


rossabaker:
Well, maybe wildcards are affected, too?

vasilmkd (approved):
This solves the issue I was seeing?

hamnis (approved):
Seems like Github Actions are reporting a hostname that is not always reachable from tests.

PR #529, opened 2021-05-07
PublishTags

### LINKED ISSUE #505: It's not publishing on tags: why not?

ChristopherDavenport:
Should fix #505

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,6 +52,12 @@ lazy val commonSettings = Seq(
 
 ThisBuild / githubWorkflowJavaVersions := Seq("adopt@1.8")
 
+ThisBuild / githubWorkflowTargetTags ++= Seq("v*")
+
+// currently only publishing tags
+ThisBuild / githubWorkflowPublishTargetBranches :=
+  Seq(RefPredicate.StartsWith(Ref.Tag("v")))
+
 ThisBuild / githubWorkflowPublishTargetBranches := Seq(
   RefPredicate.Equals(Ref.Branch("main"))
 )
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -52,6 +52,12 @@ lazy val commonSettings = Seq(
 
 ThisBuild / githubWorkflowJavaVersions := Seq("adopt@1.8")
 
+ThisBuild / githubWorkflowTargetTags ++= Seq("v*")
+
+// currently only publishing tags
+ThisBuild / githubWorkflowPublishTargetBranches :=
+  Seq(RefPredicate.StartsWith(Ref.Tag("v")))
+
```
rossabaker, line 60: These settings seem like they could go into the org plugin at some point?  Particularly in the Great Schism?

rossabaker:
I'm not sure from the regenerated workflow if this is going to do anything?

rossabaker:
Is [this](https://github.com/http4s/blaze/actions/runs/821819652/workflow#L64) the problematic line?

build.sbt
zmccoy: Absolutely unrelated to what's being done here, but what is "The Great Schism"?

rossabaker: The break between the Roman Catholic and Eastern Orthodox Church in the 11th Century.

Alternatively, [this](https://github.com/http4s/http4s/discussions/4140).

zmccoy: Thanks!  For both even

rossabaker:
I still don't understand what difference it makes from the yaml, but it matches the docs.  Maybe the plugin introspects the setting.

rossabaker:
Drat.  Didn't work.

PR #538, opened 2021-06-02
Rm sbt-dotty plugin

danicheg:
That plugin is no longer needed because of the SBT version. See https://www.scala-lang.org/blog/2021/04/08/scala-3-in-sbt.html


rossabaker (approved):
Looks like the second commit was clean, but it made me approve CI again. Running to make sure we didn't break anything.

Anyway, thanks, and :+1: on green!

isomarcte (approved):

rossabaker:
I just regenerated the workflow YAML.  I think it'll be clean after that.

PR #539, opened 2021-06-04
Use MUnit in `blaze-core` tests

danicheg:
The main goal of this PR is a unification of the test framework in `http4s`'s projects.
Changing the test framework for the `blaze-http` module it's better to make it in another PR (the diff of current PR is too big).


rossabaker:
Wow!  Were you able to do most of this with a macro or regex, or was this all manual?  This looks like a massive effort!

danicheg:
@rossabaker it would be great if was a Scalafix migration for that.. :) I've used regex for assertions and bare hands for the other stuff migration.

rossabaker:
I finally had a chance to look at this more deeply.  This is great work.

One question: why cats-effect?  There's no cats-effect in this repo, and MUnit supports `Future` results.  The `IO.fromFuture` business shouldn't be necessary.  I don't know offhand what assertion syntax plain MUnit has, but I bet you can map:

```scala
Future(6 * 7).map(f => assertEquals(_ , 42))
```

I don't know whether it supports a syntax similar to what [munit-cats-effect provides](https://github.com/typelevel/munit-cats-effect/blob/fa0867426f1a5d195ae0e562350f151bf098fc63/common/shared/src/main/scala/munit/CatsEffectAssertions.scala).  If not, we could add it in a helper, or see if the MUnit folks want it:

```scala
Future(6, 7).assertEquals(42)
```

I'm inclined to merge this, because it's a huge step in the right direction, but I think it would be more blaze-like without that extra dependency.  What do you think?

danicheg:
@rossabaker it's a definitely fairly point. I used MUnit with cats-effect mainly for assertions and thought that it might be common for all projects within http4s. But yeah, I'll try to get rid of cats-effect dependency.

danicheg:
@rossabaker and it's done.

rossabaker:
Looks good through 0e0bfa9.  I'm not clear what's going on in 46c3f00?

danicheg:
@rossabaker Thanks! The previous version of `NIO1ChannelSuite` contain `Await` and all steps in every test are in sequence. In migration on MUnit, I've migrated only assertions first. Then some tests became flaky because the logical sequence in steps was broken (e.g. the channel was closed earlier than the assertion effect fulfilled). And in 46c3f00 I've returned a logical sequence in execution.

rossabaker:
Ah, okay.  I think some of these might have been doable as `Future.successful`, but it shouldn't hurt anything to not.  And I'd like to get the move done.

PR #540, opened 2021-06-07
Add cross-compiling check to the validate command

danicheg:
It might be helpful in the local development process.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -62,7 +62,7 @@ ThisBuild / githubWorkflowPublishTargetBranches := Seq(
   RefPredicate.Equals(Ref.Branch("main"))
 )
 ThisBuild / githubWorkflowBuild := Seq(
-  WorkflowStep.Sbt(List("validate"))
+  WorkflowStep.Sbt(List("validate-ci"))
 )
 
 lazy val blaze = project.in(file("."))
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -123,4 +123,8 @@ lazy val examples = Project("blaze-examples",file("examples"))
 
 /* Helper Functions */
 
-addCommandAlias("validate", ";scalafmtCheckAll ;javafmtCheckAll ;test ;unusedCompileDependenciesTest ;mimaReportBinaryIssues")
+// use it in the local development process
+addCommandAlias("validate", ";scalafmtCheckAll ;javafmtCheckAll ;+test:compile ;test ;unusedCompileDependenciesTest ;mimaReportBinaryIssues")
+
+// use it in the CI pipeline
```
rossabaker, line 129: What is the difference between this and `validate`?  Just `test` vs. `test:compile; test`?

danicheg: Yep, here `+test:compile` to cross-compile module+tests for all Scala versions and `test` for running tests only for a single version. On CI pipeline's it's not needed because there are jobs that run for every Scala version separately.

rossabaker: I missed the `+` nuance.  What you're doing makes sense to me.

danicheg: Thanks!

PR #543, opened 2021-06-23
Use MUnit in `blaze-http` tests

danicheg:
The main goal of this PR is a unification of the test framework in `http4s`'s projects.


danicheg:
Omg, sorry about the diff 😮

danicheg:
@rossabaker if you can review this, it would be great 🙏🏻

rossabaker:
Nice work!

Confession: I looked at the beginning and end, and reviewed an arbitrary porting commit, and skimmed the rest real fast.  But what I read looks great, and I'm trusting the rest is similarly mechanical.

PR #556, opened 2021-08-29
Freshen the JDK in the Nix shell

rossabaker:
I couldn't build on a Mac with the old version.  It was missing things on SSLEngine.


rossabaker:
The build failed because 0.14.18 hasn't synced yet.  This is only used locally, and perhaps only by me.

PR #564, opened 2021-09-15
Update Scala 2.12 to `2.12.15`

danicheg:


rossabaker:
Probably I should have bumped sbt-http4s-org after releasing kind-projector, and then we wouldn't need to bump kind-projector in every repo.  Not a big deal now, but will be when we have a repo per module.

The basic tension here is that a plugin lets us maintain it in one place, but the plugin has to be updated before the projects can get the Scala bump.

PR #585, opened 2021-12-06
Update to sbt-http4s-org-0.9.0

rossabaker:


rossabaker:
This is going to make a larger build matrix.  https://github.com/http4s/sbt-http4s-org/issues/110 would be an antidote to that.

rossabaker:
Superseded by #604.

PR #600, opened 2022-01-06
Reduce level of "Accepted connection" log message

diesalbla:


rossabaker (approved):
This isn't the first time this has come up, and I've specifically targeted blaze logging configs because of it.  I think this is a more sensible default.

ChristopherDavenport (approved):

danicheg (approved):

PR #602, opened 2022-01-16
Set `githubWorkflowPublishTargetBranches` only once in the build.sbt

### LINKED ISSUE #555: Tags are not publishing from CI
https://github.com/http4s/blaze/runs/3456552938?check_suite_focus=true

*sigh*

danicheg:
This tries to fix #555. It seems that the `githubWorkflowPublishTargetBranches` should be set only once. However, I'm not sure that it's the root of the problem.


rossabaker:
From the YAML, I think this might be it.

PR #604, opened 2022-01-20
Update to sbt-http4s-org-0.10.0

rossabaker:


rossabaker:
This is a pain.  Java 9 adds [NEED_UNWRAP_AGAIN](https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLEngineResult.HandshakeStatus.html#NEED_UNWRAP_AGAIN), which should only happen under DTLS.  If we add a wildcard match, Java 8 complains about an unreachable case.  If we do nothing, Java > 8 complains about an unhandled case.  We're probably going to need to `@nowarn` our way out of this one.

rossabaker:
Failure is #376.  We weren't testing on 11 or 17 before, which is embarrassing, but we don't need to fix that in this PR.

danicheg (approved):

PR #628, opened 2022-05-20
Hello, http4s-blaze

armanbilge:
This should get things operational, but I think interesting decisions lie ahead for this project.


build.sbt
```diff
@@ -3,20 +3,40 @@ import Dependencies._
 
 val Scala212 = "2.12.15"
 val Scala213 = "2.13.8"
-val Scala3 = "3.0.2"
+val Scala3 = "3.1.2"
+val http4sVersion = "0.23.11-473-e7e64cb-SNAPSHOT"
```
armanbilge, line 7: Blaze currently depends on some unpublished changes in http4s.

build.sbt
```diff
@@ -3,20 +3,40 @@ import Dependencies._
...
+  "s01 snapshots".at("https://s01.oss.sonatype.org/content/repositories/snapshots/")
 
 ThisBuild / crossScalaVersions := Seq(Scala3, Scala212, Scala213)
 ThisBuild / scalaVersion := crossScalaVersions.value.filter(_.startsWith("2.")).last
-ThisBuild / tlBaseVersion := "0.15"
-ThisBuild / tlVersionIntroduced := Map(
-  "2.13" -> "0.14.5",
-  "3" -> "0.15.0"
-)
+ThisBuild / tlBaseVersion := "0.23"
```
armanbilge, line 15: This will bump the version for blaze's own modules, from 0.15.

build.sbt
```diff
@@ -31,24 +51,7 @@ lazy val commonSettings = Seq(
...
-      "ross@rossabaker.com",
-      url("https://github.com/rossabaker")),
-    Developer(
-      "ChristopherDavenport",
-      "Christopher Davenport",
-      "chris@christopherdavenport.tech",
-      url("https://github.com/ChristopherDavenport"))
-  ),
-  startYear := Some(2014)
+  scalafmtConfig := file(".scalafmt.blaze.conf")
```
armanbilge, line 54: I didn't have the heart to re-format the old or the new, so there are two separate configs now. I leave it to the maintainers to reconcile. I did sacrifice the examples, but anyway.

rossabaker: I'd rather reformat the old, because it moves slower than the new, but that can be done in a separate PR.

rossabaker:
I think there's more to do here eventually.  For instance, two cores is weird.  But this gets things started.

rossabaker (approved):

PR #698, opened 2022-06-12
Update `sbt-http4s-org`

danicheg:
To sum up things about Scalafix in blaze - it looks like scalafix has updated together with `RedundantSyntax` rule, which we use in old blaze modules. The diff is tiny, so I don't think we should do some workarounds here.


armanbilge:
If you're looking for a dummy rule you can use so that there is no diff, I used `NoValInForComprehension` in http4s/hpack.

I also opened:
- https://github.com/scalacenter/scalafix/issues/1613

danicheg:
this is pretty straightforward, so merging this to unlock #696

PR #729, opened 2022-08-26
Clean up Mima filters in `build.sbt`

danicheg:
Since we migrated to versioning like in `http4s`, and the `main` now relates to the `1.0.0-X` series, we can do this refactoring.


danicheg:
Alright. With the power vested in me along with blessing by CI, I merge this.

PR #730, opened 2022-08-28
Set up `mergify.yml`

danicheg:
An attempt to set up attaching those nice labels like in `http4s`.


danicheg:
@Mergifyio refresh

danicheg:
It looks like we have to merge this first to try to test (otherwise Mergify isn't applying configuration changes in this PR). So my plan is that I merge and try to test this in another PR. Hopefully, all will be good, or I'll do a revert.

PR #739, opened 2022-09-01
Fix Scala 3 updating

danicheg:
On updating Scala 3 from 3.0.x to 3.1.x we forgot to update the Scala Steward config. So we are stuck on Scala 3.1.2 for a while. Let's fix this.


danicheg:
Actually, there isn't rocket science here. So let's go onward.

PR #747, opened 2022-09-19
Merge `series/0.23` into `main`

danicheg:


danicheg:
Oops. Apparently, it's too early to do this job. The core module should release the fresh milestone first.

PR #776, opened 2022-11-30
Mark "split large buffers" test flaky

rossabaker:
Alternative to #771.  Better discussion there, but this reflects that we don't yet understand it, but also lets us move forward.

### DIFF
--- a/core/src/test/scala/org/http4s/blaze/pipeline/stages/SSLStageSuite.scala
+++ b/core/src/test/scala/org/http4s/blaze/pipeline/stages/SSLStageSuite.scala
@@ -61,7 +61,7 @@ class SSLStageSuite extends BlazeTestSuite {
       } yield ()
     }
 
-    test(testSuitePrefix + " should split large buffers".flaky) {
+    test((testSuitePrefix + " should split large buffers").flaky) {
       // See https://github.com/http4s/blaze/pull/771 for a deeper
       // analysis of the flake and possible solutions.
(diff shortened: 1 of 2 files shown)


rossabaker:
Uh... does `.flaky` not do what I remember?

core/src/test/scala/org/http4s/blaze/pipeline/stages/SSLStageSuite.scala
```diff
@@ -61,7 +61,9 @@ class SSLStageSuite extends BlazeTestSuite {
       } yield ()
     }
 
-    test(testSuitePrefix + " should split large buffers") {
+    test((testSuitePrefix + " should split large buffers").flaky) {
```
danicheg, line 64: It makes me whaaaat
```scala
test(testSuitePrefix + " should split large buffers".flaky) {} // compiles but doesn't work properly
```

danicheg:
@rossabaker hope you don't mind that I dropped my few lines here

core/src/test/scala/org/http4s/blaze/pipeline/stages/SSLStageSuite.scala
danicheg: though nothing new, silly type-weakness of string concatenation

rossabaker: Oh, good find.  I always thought it was odd that operator went on the String in the first place.

PR #789, opened 2023-02-03
Minimize logger creation impact in IdentityWriter

plokhotnyuk:
Currently for 0.23.13 version `IdentityWriter` is burning more CPU cycles in logger creation than in using to render responses:
image

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blaze/core/util/IdentityWriter.scala
+++ b/blaze-core/src/main/scala/org/http4s/blaze/core/util/IdentityWriter.scala
@@ -30,7 +30,6 @@ private[blaze] class IdentityWriter[F[_]](size: Long, out: TailStage[ByteBuffer]
     protected val F: Async[F]
 ) extends Http1Writer[F] {
 
-  private[this] val logger = getLogger
   private[this] var headers: ByteBuffer = null
 
   private var bodyBytesWritten = 0L
@@ -50,7 +49,7 @@ private[blaze] class IdentityWriter[F[_]](size: Long, out: TailStage[ByteBuffer]
       val msg =
         s"Will not write more bytes than what was indicated by the Content-Length header ($size)"
 
-      logger.warn(msg)
+      IdentityWriter.logger.warn(msg)
 
       val reducedChunk = chunk.take((size - bodyBytesWritten).toInt)
       writeBodyChunk(reducedChunk, flush = true).flatMap(_ =>
@@ -78,11 +77,15 @@ private[blaze] class IdentityWriter[F[_]](size: Long, out: TailStage[ByteBuffer]
     else {
       val msg = s"Expected `Content-Length: $size` bytes, but only $total were written."
 
-      logger.warn(msg)
+      IdentityWriter.logger.warn(msg)
 
       writeBodyChunk(chunk, flush = true).flatMap(_ =>
         Future.failed(new IllegalStateException(msg))
       )(parasitic)
     }
   }
+}
+
+private[blaze] object IdentityWriter {
+  private lazy val logger = getLogger
 }


blaze-core/src/main/scala/org/http4s/blaze/core/util/IdentityWriter.scala
```diff
@@ -78,11 +77,15 @@ private[blaze] class IdentityWriter[F[_]](size: Long, out: TailStage[ByteBuffer]
...
 
       writeBodyChunk(chunk, flush = true).flatMap(_ =>
         Future.failed(new IllegalStateException(msg))
       )(parasitic)
     }
   }
 }
+
+private[blaze] object IdentityWriter {
+  private lazy val logger = getLogger
```
vilunov, line 90: minor nitpick: outer object is lazy already, you can make this val a non-lazy one

Baccata: In my experience, you should never eagerly compute an expression in a static object if that expression contains any kind of side effect, which is the case here. 

By making it lazy, you have a chance to witness potential exceptions as they'll bubble up to your main method. Making it eager puts you at risk of a hard-to-debug deadlock that'll just keep your JVM hanging. 

Now that being said, maybe there's something about making static objects private that changes the behaviour I'm describing.

danicheg: Agree on that one with @Baccata

danicheg:
@plokhotnyuk just wondering, have you tried to create an updated flame graph with this change?

blaze-core/src/main/scala/org/http4s/blaze/core/util/IdentityWriter.scala
plokhotnyuk: From other size lazy vals deffer initialization that can introduce latency spikes especially under contention pressure at max load.

Also, they not so stable these days: https://twitter.com/SimonrDev/status/1621516492482433032

IMHO a better option would be replacing all logger instances by one in some utility object that is initialized at startup.

plokhotnyuk:
> @plokhotnyuk just wondering, have you tried to create an updated flame graph with this change?

I will if it help for faster merge and release

blaze-core/src/main/scala/org/http4s/blaze/core/util/IdentityWriter.scala
rossabaker: Any instance loggers might benefit by moving to the companion object.  The cost would be a few femtoseconds on the static method call instead of a `private[this]` field reference, but allocations would usually make this a win.

Consolidating to one global logger would put all log statements into one category.  They still need to be instantiated with the class or companion they're relevant to.  (Or multiple put in one object with explicit names, but that would be an endless source of copy-and-paste errors.)

rossabaker (approved):

SethTisue:
I don't know if it's _necessary_, but it would certainly be _interesting_ to see the "after" flame graph?

danicheg:
Exactly, Seth, you just have decomposed what I meant by 'wondering' above.

bryce-anderson:
First of all, I'm fine with moving the logger to the object.

That said, is my intuition wrong or is the performance profile surprising? In general, I never considered getting a logger to be a heavyweight operation. Certainly not significantly more expensive as rendering the headers and body.

blaze-core/src/main/scala/org/http4s/blaze/core/util/IdentityWriter.scala
vilunov: To be honest, I can't really imagine a scenario where an slf4j logger initialization would result in a JVM deadlock. It would cause a fatal "exception in initializer" error in the worst case (which I never had even in extreme experiments), and it would even have a stack trace, but not a deadlock.

By the way, most of Java ecosystem actually relies on putting logger instances in a static value, not even a static method, so if we really started having such a problem with logger inits, then we are in for a very fun ride indeed. Generally, putting side-effects in even lazy vals in companion objects is not a very robust practice, but slf4j is so prevalent and battle-tested (more so than lazy vals themselves to be honest), so I doubt it will have a major impact on debugging experience.

rossabaker: I'm not convinced the lazy makes a real difference either way here, but:
* `getLogger` is a macro that expands to the traditional Java call, which as @vilunov notes, is usually a private static final.
* The log4s documentation uses eager vals in objects.
* They're mostly (all?) eager in the rest of Blaze.

rossabaker:
I'm well convinced that this is a positive change, but I'm skeptical it will make a drastic difference in production.

1. Firstly, I don't see anything writing bytes.  Is this an some in-memory test?  I can't piece together exactly what slice of functionality is under test here.
2. Secondly, it is curious to me that log4j is doing a stack walk, after the log4s macro already infers the class.  I usually use logback, and would be curious if that generates a similar flame graph before this refactor.

I'm still in favor of this.  I just don't think anybody is going to cut 20% from their cloud bill because of it.

plokhotnyuk:
The binaries which built from the main branch do not work for me due to binary incompatibilities.

I'm going to apply the patch for the last commit related to 0.23 series.

How can I find that commit?

<details>
  <summary>Here are lists of remote branches and tags after cloning of https://github.com/plokhotnyuk/blaze</summary>

```
andriy@notebook:~/Projects/com/github/plokhotnyuk/blaze$ git branch -r
  origin/HEAD -> origin/master
  origin/Http2HeaderCodec
  origin/build-info
  origin/client-core-improvements
  origin/exposeSerialExecutor
  origin/faster-nio1
  origin/issue/175_scalastyle
  origin/issue/186-Selectable
  origin/issue74
  origin/main
  origin/master
  origin/memoryleak
  origin/patch-1
  origin/release-0.12.x
  origin/release-0.13.x
  origin/removeDisconnected
  origin/rossabaker-patch-1
  origin/sslInfinteloop
  origin/stages-improvements
  origin/tmp/echo-benchmark
  origin/topic/http2cleanup
  origin/topic/in-n-out
  origin/topic/javacOptions
  origin/topic/scala-2.12
  origin/updateSbtRig
  origin/websocket-0.2.1
  origin/wip/responder
  origin/wip/wsclient
andriy@notebook:~/Projects/com/github/plokhotnyuk/blaze$ git tag -l
v0.1.0
v0.10.0
v0.10.1
v0.11.0
v0.12.0
v0.12.1
v0.12.10
v0.12.11
v0.12.12
v0.12.13
v0.12.2
v0.12.3
v0.12.4
v0.12.5
v0.12.6
v0.12.7
v0.12.8
v0.12.9
v0.13.0
v0.14.0-M1
v0.14.0-M2
v0.14.0-M3
v0.2.0
v0.3.0
v0.4.0
v0.4.1
v0.5.0
v0.5.1
v0.6.0
v0.7.0
v0.8.0
v0.8.1
v0.8.2
v0.9.0

```
</details>

rossabaker:
The series/0.23 branch is compatible with the latest http4s-core-0.23.

rossabaker:
Incorporated via #792.

Issue #772, opened 2022-11-20
Changes to `async_` cancelation semantics


armanbilge:
See:
- https://github.com/typelevel/cats-effect/pull/3205

armanbilge:
I let CI try this in https://github.com/armanbilge/blaze/commit/df420f24ce4b0dd9a541f7b8536ca778281a6f36 and there are some test failures.

```
[error] Failed tests:
[error] 	org.http4s.blaze.client.PoolManagerSuite
[error] 	org.http4s.blaze.client.ClientTimeoutSuite
[error] 	org.http4s.blaze.client.BlazeClientSuite
```

armanbilge:
Release candidate is available.
- https://github.com/typelevel/cats-effect/releases/tag/v3.5.0-RC1

rossabaker:
Most of the `async_` calls don't have obvious resources to release.  There's the stage, but shutdown of the stage is probably why we're being canceled.  My two kneejerk reactions are the `Some(F.unit).pure[F]` that's generally discouraged, or shutting down the stage again just in case.  I'm just not sure.

armanbilge:
`Some(F.unit).pure[F]` seems strange, sometimes what I've seen is `F.delay(setupCallback()).as(Some(F.unit))`

rossabaker:
Oh, right.  `async_` hides the delay of setting up the callback.  It should look more like your version if we go that route.

rossabaker:
Addressed in #806

PR #791, opened 2023-02-20
cats-effect-3.5

rossabaker:
The basic strategy is to shut down a stage if cancellation happens during an IO operation.  Cancellation probably happens because we're already shutting down the stage, but this makes sure.

Getting ahead of this.  See #772.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Support.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Support.scala
@@ -76,13 +76,17 @@ private final class Http1Support[F[_]](
     connectTimeout,
   )
 
+  /* This is an uncancelable operation, and it's the caller's
+   * responsibility to guarantee this connection gets closed.
+   */
   def makeClient(requestKey: RequestKey): F[BlazeConnection[F]] =
     getAddress(requestKey) match {
       case Right(a) =>
         fromFutureNoShift(
           executionContextConfig.getExecutionContext.flatMap(ec =>
             F.delay(buildPipeline(requestKey, a, ec))
-          )
+          ),
+          None,
         )
       case Left(t) => F.raiseError(t)
     }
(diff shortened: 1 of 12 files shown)


rossabaker:
All the tests passed locally without any changes to `async_`.  That doesn't mean there shouldn't be any, but that's not what was observed on #772 with an earlier snapshot.

armanbilge:
Huh, well that's good news then? 🤔 thanks for looking into this.

Note that the change to `async_` also affects stuff built on it, like `fromFuture`.

rossabaker:
We don't use `fromFuture`, but have a `fromFutureNoShift` that I think dates back to Cats-Effect 2 that embeds one of these.  Blargh.

armanbilge:
> but that's not what was observed on #772 with an earlier snapshot.

I think I committed that test against the `main` branch, not `series/0.23`. Not really sure why it would make such a major difference though.

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Support.scala
```diff
@@ -76,13 +76,17 @@ private final class Http1Support[F[_]](
...
+   */
   def makeClient(requestKey: RequestKey): F[BlazeConnection[F]] =
     getAddress(requestKey) match {
       case Right(a) =>
         fromFutureNoShift(
           executionContextConfig.getExecutionContext.flatMap(ec =>
             F.delay(buildPipeline(requestKey, a, ec))
-          )
+          ),
+          None,
```
rossabaker, line 89: I'd like to make this method return a resource, but it goes straight into the pool, and refactoring that will be a rippling change.

rossabaker (commented):
I'm not really sure how to go about testing these changes.  Failing tests where I/O operations are canceled would be ideal.

rossabaker:
Seems as though I've done real damage to the PoolManager with this.

rossabaker:
Obsolete and wrong.

PR #798, opened 2023-03-19
Remove `util.Execution.withTimeout`

danicheg:
It's package private and unused overall.


danicheg:
Ah, yes, Mima wouldn't like this. Perhaps this makes sense to 1.0.

rossabaker:
Can be deprecated here to make sure it doesn't spread.

PR #806, opened 2023-05-14
Upgrade to http4s-0.23.19

rossabaker:
This is a fairly urgent fix after cats-effect-3.5.0.

* Fixes deprecation warnings from fs2-3.7.0
* Addresses changes to semantics of `F.async_`.  The general strategy is to shutdown the stage.  Many of the changes are whitespace.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
@@ -179,8 +179,6 @@ private final class Http1Connection[F[_]](
 
   override protected def contentComplete(): Boolean = parser.contentComplete()
 
-  private[this] val noopCancel = Some(F.unit)
-
   private def executeRequest(
       req: Request[F],
       cancellation: F[TimeoutException],
@@ -218,7 +216,7 @@ private final class Http1Connection[F[_]](
           val idleTimeoutF: F[TimeoutException] = idleTimeoutStage match {
             case Some(stage) =>
               F.async[TimeoutException] { cb =>
-                F.delay(stage.setTimeout(cb)).as(noopCancel)
+                F.delay(stage.setTimeout(cb)).as(Some(F.delay(stage.cancelTimeout())))
               }
             case None => F.never[TimeoutException]
           }
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
@@ -179,6 +179,8 @@ private final class Http1Connection[F[_]](
 
   override protected def contentComplete(): Boolean = parser.contentComplete()
 
+  private[this] val noopCancel = Some(F.unit)
+
   private def executeRequest(
       req: Request[F],
       cancellation: F[TimeoutException],
@@ -214,7 +216,10 @@ private final class Http1Connection[F[_]](
             }
 
           val idleTimeoutF: F[TimeoutException] = idleTimeoutStage match {
-            case Some(stage) => F.async_[TimeoutException](stage.setTimeout)
+            case Some(stage) =>
+              F.async[TimeoutException] { cb =>
+                F.delay(stage.setTimeout(cb)).as(noopCancel)
+              }
             case None => F.never[TimeoutException]
           }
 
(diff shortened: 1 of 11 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
```diff
@@ -203,75 +205,77 @@ private final class PoolManager[F[_], A <: Connection[F]](
...
-                  createConnection(key, callback)
-
-              case None if maxConnectionsPerRequestKey(key) <= 0 =>
-                F.delay(callback(Left(NoConnectionAllowedException(key))))
-
-              case None if curTotal == maxTotal =>
-                val keys = idleQueues.keys
-                if (keys.nonEmpty)
+      semaphore.permit
+        .use { _ =>
```
armanbilge, line 209:
```suggestion
      semaphore.permit
        .surround {
```

blaze-core/src/main/scala/org/http4s/blazecore/util/package.scala
```diff
@@ -42,6 +42,8 @@ package object util extends ParasiticExecutionContextCompat {
         case Some(value) =>
           F.fromTry(value)
         case None =>
+          // Scala futures are uncancelable.  There's not much we can
+          // do here other than async_.
```
armanbilge, line 46: Or leak, in some bad situations 🤷 it's a crappy api

armanbilge (approved):
Don't know anything about Blaze but this looks reasonable.

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -214,7 +216,10 @@ private final class Http1Connection[F[_]](
             }
 
           val idleTimeoutF: F[TimeoutException] = idleTimeoutStage match {
-            case Some(stage) => F.async_[TimeoutException](stage.setTimeout)
+            case Some(stage) =>
+              F.async[TimeoutException] { cb =>
+                F.delay(stage.setTimeout(cb)).as(noopCancel)
```
armanbilge, line 221: Hm, is `cancelTimeout()` not the right thing to do here?
https://github.com/http4s/blaze/blob/8cf873e228c89b8ed836d7c635ab8142222ebc85/blaze-core/src/main/scala/org/http4s/blaze/core/IdleTimeoutStage.scala#L134

blaze-client/src/main/scala/org/http4s/blaze/client/PoolManager.scala
rossabaker: I found one of these somewhere else!

blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
rossabaker: That seems legit.  In practice, I think the stage is going to shutdown before any of this matters.

danicheg (approved):
It'd be nice to test it on somebody's production 😅

rossabaker:
Yeah, this needed an RC before the CE breakage hit production.  We can follow up with another patch if this was not the way, but shit's already broke.

rossabaker:
It's a holiday in the US, so my time is limited, but I will try to get this released tonight.

danicheg:
I can draft release notes if this helps.

rossabaker:
That would be wonderful, thanks!

danicheg:
Good, here we go https://github.com/http4s/blaze/releases/edit/untagged-d23cbac392ab1cfa87d4. Look at it before publishing, I might have written some nonsense 🙏🏻

rossabaker:
Those look great, thanks.  Publishing now.

rossabaker:
A bunch of tests failed on release.  Perhaps shutting down the stages was too aggressive.  *sigh*

rossabaker:
I thought it might be that the channel.addShutdownHook doesn't backpressure and rewrote the tests as a latch ... but according to the comment, it should.  Something is stopping those shutdown hooks from running reliably, but I can't reproduce it locally.

rossabaker:
I can't see what that test has to do with these changes.  I'm rerunning a few times, and I might just try to publish the tag again.

rossabaker:
I can't reproduce the failures that slew v0.23.15 after multiple attempts locally or in CI.  I'm just going to try the tag again.

PR #807, opened 2023-05-14
Pin scalafmt to 3.6.x

rossabaker:
As discussed in #786 and brought back to fore by #800.  Can unpin this if someone wants to dig into why our trailing commas keep disappearing.


rossabaker:
I think this build has gone space truckin'.  We changed some cancellation semantics in #806, but this shouldn't be affected by cats-effect-3.5.0, which is what that fixes.  :shrug:

danicheg (approved):

rossabaker:
Since this hang is from before the cancellation semantics changed, I don't know that it's meaningful.  We'll need to keep an eye on it, but this is just a pin, so, merging.

PR #849, opened 2024-01-15
Provide a custom way to handle SSL exceptions

arturaz:
Same thing that was provided for ember (https://github.com/http4s/http4s/pull/7093).

### DIFF
--- a/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
+++ b/blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
@@ -31,7 +31,7 @@ import com.comcast.ip4s.SocketAddress
 import org.http4s.blaze.channel._
 import org.http4s.blaze.channel.nio1.NIO1SocketServerGroup
 import org.http4s.blaze.pipeline.LeafBuilder
-import org.http4s.blaze.pipeline.stages.SSLStage
+import org.http4s.blaze.pipeline.stages.{SSLStage, SSLStageDefaults}
 import org.http4s.blaze.server.BlazeServerBuilder._
 import org.http4s.blaze.util.TickWheelExecutor
 import org.http4s.blaze.{BuildInfo => BlazeBuildInfo}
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -54,10 +54,30 @@ private object SSLStage {
...
+  */
+object SSLStageDefaults {
+
+  /** Default value for [[SSLStage.maxWrite]] */
+  final val MaxWrite = 1024 * 1024
+}
+
+/** @param maxWrite
+  *   \@see [[SSLStageDefaults.MaxWrite]].
+  */
```
danicheg, line 69: It seems like missing scaladoc

core/src/main/scala/org/http4s/blaze/pipeline/stages/SSLStage.scala
```diff
@@ -54,10 +54,30 @@ private object SSLStage {
...
+final class SSLStage(
+    engine: SSLEngine,
+    maxWrite: Int,
+    sslHandshakeExceptionHandler: PartialFunction[Throwable, Unit]
```
danicheg, line 73: If it's expected to handle `SSLException`s only, shouldn't the contract be `PF[SSLException, Unit]`?

blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
```diff
@@ -194,15 +194,37 @@ class BlazeServerBuilder[F[_]] private (
...
     copy(sslConfig = new ContextWithClientAuth[F](sslContext, clientAuth))
 
+  /** Configures the server with TLS, using the provided `SSLContext` and its
+    * default `SSLParameters`
+    *
+    * @param sslErrorHandler function that runs if an error occurs during the TLS handshake. Default behavior is to log the error.
+    */
+  def withSslContext(
+      sslContext: SSLContext,
+      sslErrorHandler: PartialFunction[Throwable, Unit] = PartialFunction.empty,
```
danicheg, line 204: Wouldn't it be more beneficial to have a general error handler exposed to users and not only for SSL exceptions?

danicheg (commented):
This seems like a good addition to Blaze, but I suggest we consult Ross before making any decisions.

blaze-server/src/main/scala/org/http4s/blaze/server/BlazeServerBuilder.scala
danicheg: IIUC, Ember exposes a general error handler? https://github.com/http4s/http4s/blob/series/0.23/ember-server/shared/src/main/scala/org/http4s/ember/server/EmberServerBuilder.scala#L46-L47

rossabaker: For what it's worth, Ember's general error handler just came up in https://github.com/http4s/http4s/pull/7399.  This is more in the direction that Ember is going, but probably away from the direction I'd prefer to go.

PR #857, opened 2024-03-04
Switch to a Nix flake

rossabaker:
That shell is quite archaic.


rossabaker:
This isn't production code.  Merging with an iron fist.

PR #913, opened 2024-10-24
Implement Connection persistence per RFC 9112, Section 9.3

rossabaker:
We are aggressively closing connections in the presence of `Connection` headers composed exclusively of options other than `keep-alive` or `close`.  This is contrary to [the RFC](https://datatracker.ietf.org/doc/html/rfc9112#section-9.3)

### DIFF
--- a/blaze-core/src/main/scala/org/http4s/blaze/core/Http1Stage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blaze/core/Http1Stage.scala
@@ -26,6 +26,7 @@ import org.http4s.Entity.Empty
 import org.http4s.Header
 import org.http4s.Header.Raw
 import org.http4s.Headers
+import org.http4s.HttpVersion
 import org.http4s.InvalidBodyException
 import org.http4s.Method
 import org.http4s.Request
@@ -67,6 +68,7 @@ private[blaze] trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
   protected def contentComplete(): Boolean
 
   /** Check Connection header and add applicable headers to response */
+  @deprecated("Use checkRequestCloseConnection(Request) instead", "0.23.17")
   protected final def checkCloseConnection(conn: Connection, rr: StringWriter): Boolean =
     if (conn.hasKeepAlive) { // connection, look to the request
       logger.trace("Found Keep-Alive header")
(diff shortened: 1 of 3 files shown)


blaze-core/src/main/scala/org/http4s/blaze/core/Http1Stage.scala
```diff
@@ -83,6 +85,32 @@ private[blaze] trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
       true
     }
 
+  private[http4s] final def checkRequestCloseConnection(req: Request[F]): Boolean = {
```
rossabaker, line 88: I tried to fix the Boolean blindness, but it ripples to many places.

blaze-core/src/main/scala/org/http4s/blaze/core/Http1Stage.scala
```diff
@@ -67,6 +68,7 @@ private[blaze] trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
   protected def contentComplete(): Boolean
 
   /** Check Connection header and add applicable headers to response */
+  @deprecated("Use checkRequestCloseConnection(Request) instead", "0.23.17")
```
rossabaker, line 71: Kept for binary compatibility, though we could do the MiMa dance.

blaze-server/src/main/scala/org/http4s/blaze/server/Http1ServerStage.scala
```diff
@@ -274,10 +269,14 @@ private[blaze] class Http1ServerStage[F[_]](
             case _ => // nop
           }
 
-        // add KeepAlive to Http 1.0 responses if the header isn't already present
-        rr << (if (!closeOnFinish && parser.minorVersion() == 0 && respConn.isEmpty)
-                 "Connection: keep-alive\r\n\r\n"
-               else "\r\n")
+        closeOnFinish match {
+          case true if respConn.isEmpty =>
+            rr << "Connection: close\r\n\r\n"
```
rossabaker, line 274: This was a buried side effect in the old code.

rossabaker (commented):
Sharing for early feedback on the approach before I figure out how to test it.

milessabin:
LGTM.

Maybe a bit off-topic for this PR, but when returning a `Connection: close` do you pause before closing to allow the client to do an active close avoiding `TIME_WAIT` on the server side?

rossabaker:
Hell, I started from the wrong branch.

rossabaker:
Miles has a good question that I haven't dug into yet.  Spinning that off into an issue.

PR #917, opened 2024-11-02
Fix Connection persistence according to RFC 9112

rossabaker:
We are aggressively closing connections in the presence of `Connection` headers composed exclusively of options other than `keep-alive` or `close`.  This is contrary to [the RFC](https://datatracker.ietf.org/doc/html/rfc9112#section-9.3)

- `Connection` tokens other than `close` and `keep-alive` are no longer considered in persistence logic.
- blaze-server will no longer add an implicit `Connection: close` header to the response unless required by RFC 9112.
- blaze-client will no longer add a duplicate `Connection: close` header when the request already has one.

Replaces #913, which targeted the incorrect branch.

### DIFF
--- a/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
+++ b/blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
@@ -32,6 +32,7 @@ import org.http4s.blaze.pipeline.Command.EOF
 import org.http4s.blazecore.Http1Stage
 import org.http4s.blazecore.IdleTimeoutStage
 import org.http4s.blazecore.util.Http1Writer
+import org.http4s.blazecore.util.NullWriter
 import org.http4s.client.RequestKey
 import org.http4s.headers.Host
 import org.http4s.headers.`Content-Length`
@@ -200,10 +201,8 @@ private final class Http1Connection[F[_]](
           if (userAgent.nonEmpty && !req.headers.contains[`User-Agent`])
             rr << userAgent.get << "\r\n"
 
-          val mustClose: Boolean = req.headers.get[HConnection] match {
-            case Some(conn) => checkCloseConnection(conn, rr)
-            case None => getHttpMinor(req) == 0
-          }
+          val mustClose: Boolean =
+            checkRequestCloseConnection(req.headers.get[HConnection], getHttpMinor(req), NullWriter)
 
           val writeRequest: F[Boolean] = getChunkEncoder(req, mustClose, rr)
             .write(rr, req.body)
--- a/blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
+++ b/blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
@@ -60,6 +60,7 @@ private[http4s] trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
   protected def contentComplete(): Boolean
 
   /** Check Connection header and add applicable headers to response */
+  @deprecated("Use checkConnectionPersistence(Option[Connection], Int, Writer) instead", "0.23.17")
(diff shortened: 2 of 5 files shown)


blaze-client/src/main/scala/org/http4s/blaze/client/Http1Connection.scala
```diff
@@ -200,10 +201,8 @@ private final class Http1Connection[F[_]](
           if (userAgent.nonEmpty && !req.headers.contains[`User-Agent`])
             rr << userAgent.get << "\r\n"
 
-          val mustClose: Boolean = req.headers.get[HConnection] match {
-            case Some(conn) => checkCloseConnection(conn, rr)
-            case None => getHttpMinor(req) == 0
-          }
+          val mustClose: Boolean =
+            checkRequestCloseConnection(req.headers.get[HConnection], getHttpMinor(req), NullWriter)
```
rossabaker, line 205: I had a hard time teasing apart the side effect of writing the `Connection: close` header, but we don't even want the side effect on the client side.

Alternatives would be to:
- duplicate the connection persistence logic
- put a protected method in `Http1Stage` to represent the side effect, or lack thereof
- wrap this writer in an Option
- question all our lives' choices

blaze-core/src/main/scala/org/http4s/blazecore/Http1Stage.scala
```diff
@@ -76,6 +77,42 @@ private[http4s] trait Http1Stage[F[_]] { self: TailStage[ByteBuffer] =>
...
+  ): Boolean =
+    if (conn.fold(false)(_.hasClose)) {
+      logger.trace(s"Closing ${conn} due to explicit close option in request's Connection header")
+      // This side effect doesn't really belong here, but is relied
+      // upon in multiple places as we look for the encoder.  The
+      // related problem of writing the keep-alive header in HTTP/1.0
+      // is handled elsewhere.
+      if (minorVersion >= 1) {
+        rr << "Connection: close\r\n"
+      }
```
rossabaker, line 101: I tried to replace the `Boolean` with a sum type to include whether the close is implicit or explicit.  But the result threads through fairly deeply as a `Boolean`.

blaze-server/src/test/scala/org/http4s/blaze/server/ServerTestRoutes.scala
```diff
@@ -57,13 +57,19 @@ object ServerTestRoutes extends Http4sDsl[IO] {
       // ///////////////////////////////
       (
         "GET /get HTTP/1.0\r\nConnection:close\r\n\r\n",
-        (Status.Ok, Set(length(3), textPlain, connClose), "get"),
+        (Status.Ok, Set(length(3), textPlain), "get"),
```
rossabaker, line 60: All removals of `connClose` from this test are because it's the default in HTTP/1.0.  There's no need for us to reiterate that.

blaze-server/src/test/scala/org/http4s/blaze/server/ServerTestRoutes.scala
```diff
@@ -57,13 +57,19 @@ object ServerTestRoutes extends Http4sDsl[IO] {
...
       (
         "GET /get HTTP/1.1\r\nConnection:close\r\n\r\n",
         (Status.Ok, Set(length(3), textPlain, connClose), "get"),
       ),
+      // ///////////////////////////////
+      // Don't close connection on an unrecognized Connection header
+      (
+        "GET /get HTTP/1.1\r\nConnection: fiddle-faddle\r\n\r\n",
+        (Status.Ok, Set(length(3), textPlain), "get"),
+      ),
```
rossabaker, line 72: This test exposes the behavior change that motivated this PR.

rossabaker:
Merging on Miles' review.

